PR #21, opened 2025-05-28
ENH: Build React app on pip install

### LINKED ISSUE #1: Replace `static/index.html` with a React "Hello world"
It's not clear what the build / distribution process should be. However, the compiled React app should be distributed in the Python package's `static/` subpackage so there is likely a build step that builds it into this directory. Otherwise it could be part of a CI step, which is probably necessary to ensure it's up-to-date either way.

mferrera:
Resolves #1

## Checklist

- [x] Tests added (if not, comment why)

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -7,6 +7,7 @@
     "dev": "vite",
     "build": "tsc -b && vite build --outDir ../src/fmu_settings_gui/static --emptyOutDir",
     "lint": "tools/biome check && eslint .",
+    "ci-lint": "biome ci && eslint .",
     "preview": "vite preview",
     "openapi-ts": "openapi-ts"
   },
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ on:
     - cron: "0 0 * * *"
 
 jobs:
-  checkv:
+  check:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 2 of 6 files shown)


frontend/package.json
```diff
@@ -7,6 +7,7 @@
     "dev": "vite",
     "build": "tsc -b && vite build --outDir ../src/fmu_settings_gui/static --emptyOutDir",
     "lint": "tools/biome check && eslint .",
+    "ci-lint": "biome ci && eslint .",
```
joargr, line 10: Based on the TODO and commented command in `.github/workflows/publish.yml` it seems that the `ci-lint` script isn't used as of now. It could be a bit confusing to see the `biome` command invoked from the `tools` subdirectory for the `lint` script but invoked from the base directory for the `ci-lint` script, but I guess that's due to the two different ways the `biome` tool is installed (downloaded for the manual installation to TGX node and through workflow action in the workflow run to GitHub runner). I can add a note about this in the `frontend/README.md` file.

joargr (approved):
I haven't tried running this but it looks good.

frontend/package.json
joargr: I don't think that the code currently in the main branch produces errors or warnings from biome or the type-checked linting. It seems that no matter what the current directory is when Biome is started, it will check the files in the same directories, as specified in `files.includes` in the `biome.json` configuration file. As an aside, this configuration is in the root of the repo instead of in the `frontend` directory, as the Biome Visual Studio Code plugin otherwise won't find the file.

PR #33, opened 2025-06-27
MAINT: Have query/mutation and form logic separate

### LINKED ISSUE #32: Move API key mutation out from common EditableTextFieldForm
The `EditableTextFieldForm` component should be common and not tied to a specific data type or mutation. Currently the configuration of the form is tailored to API keys and uses their mutation. The code should be refactored such that this component can be common and used for other kinds of data as well.

joargr:
Resolves #32

Adds a component `KeysTextFieldForm` for setting up data query and mutation, which uses the `EditableTextFieldForm` for form logic. Uses some callback functions to let the two major operations do the necessary logic.

When the user submits the form, the following happens:
1. The form's `onSubmit` function executes, which calls the `mutationCallback` defined in the query/mutation component
2. The `mutationCallback` function executes, which performs the mutation (saving of data to the API). A successful mutation will call the `formSubmitCallback` function, defined in the form component
3. The `formSubmitCallback` function executes, which will show a toast message to the user and reset the form, so that the form will show the updated value
4. If the mutation results in an error, a toast error message will be shown to the user. This is handled by the default error handling of the query/mutation client

See also description on some of the code in the files.

## Checklist

### DIFF
--- a/frontend/src/components/form.tsx
+++ b/frontend/src/components/form.tsx
@@ -8,11 +8,6 @@ import {
 import { error_filled } from "@equinor/eds-icons";
 import { createFormHook } from "@tanstack/react-form";
 import {
-  QueryClient,
-  useMutation,
-  useSuspenseQuery,
-} from "@tanstack/react-query";
-import {
   ChangeEvent,
   Dispatch,
   SetStateAction,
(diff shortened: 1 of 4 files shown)


frontend/src/routes/user/keys.tsx
```diff
@@ -1,16 +1,96 @@
...
+    formReset,
+  }: MutationCallbackProps<StringObject>) => {
+    mutate(
+      { body: { id: name, key: formValue[name] } },
+      {
+        onSuccess: (data) => {
+          formSubmitCallback({
+            message: data.message,
+            formReset,
+          });
```
joargr, line 74: At mutation success the function for completing the form submit handling is called. Note that the function to reset the form is passed back to this function, where it will be executed.

frontend/src/components/form.tsx
```diff
@@ -117,67 +132,55 @@ const { useAppForm: useAppFormEditableTextField } = createFormHook({
...
 
+  const formSubmitCallback = ({
+    message,
+    formReset,
+  }: FormSubmitCallbackProps) => {
+    toast.info(message);
+    formReset();
+    setIsReadonly(true);
+  };
+
```
joargr, line 173: This function will be executed at a successful mutation, giving feedback to the user and resetting the form. If the mutation results in an error, the function will not be executed, leaving the form in the edit mode. The user can use the cancel button to go back to the original value. A toast error message will also be shown, triggered by the default error handling.

frontend/src/components/form.tsx
```diff
@@ -22,18 +17,38 @@ import {
...
+interface FormSubmitCallbackProps {
+  message: string;
+  formReset: () => void;
+}
+
+export interface MutationCallbackProps<T> {
+  formValue: T;
+  formSubmitCallback: (props: FormSubmitCallbackProps) => void;
+  formReset: () => void;
+}
```
joargr, line 36: These are properties concerning mutation handling. The use of generics for the type of `formValue` is in anticipation of other forms, where the data processed in the form might have different types.

PR #38, opened 2025-07-31
Add SSO access token to session

### LINKED ISSUE #34: Add SSO access token to API session
The user's SSO access token should be added to the API session, so that the API can use it for querying the SMDA API.

joargr:
Resolves #34 

Lets the user log in to the Equinor SSO, and takes the resultant access token and adds it to the API session. When the user goes to the SMDA page, an SMDA health check is performed in the API, showing whether the prerequisites are met for the API to communicate with the SMDA API. These prerequisites are:
- An SMDA subscription key
- An SSO access token

If the health check fails, each prerequisite is checked by the page, and information is shown to the user on how to fix the missing data.

- Uses the MSAL.js library
- Uses redirect for going to the login page and not popup, as there would likely be issues with popup blocking
- The `clientId` (for fmu-settings) and `tenantId` (for Equinor) are stored in the `config.ts` file. These values are not considered to be secrets, so there should be no problems with storing them in the repo
- The response from the SMDA health check is converted before being used in the SMDA page. A 401 Unauthorized response from the API is changed into a boolean `false` value, to make it better fit with how the query library functions (ie. such a response won't be treated as an error)
- The access token is automatically added to the session, also when the session has expired and a new session is created
- If for some reason the API has an outdated access token, the checks on the SMDA page will enable the user to manually add the current access token to the session
- The MSAL acquire token function will if necessary do a refresh, meaning it will use a refresh token to get a new access token. This happens automatically in the background. If necessary, a redirect to the login page will also happen, thus the user will in some cases experience being redirected to a login page when accessing the SMDA page (ie. initiating a login without pressing the login button)
- The endpoint functions have been renamed in the API, so the GUI code has been updated to reflect this

## Checklist

### DIFF
--- a/frontend/src/main.tsx
+++ b/frontend/src/main.tsx
@@ -76,6 +76,8 @@ declare module "@tanstack/react-query" {
   }
 }
 
+const msalInstance = new PublicClientApplication(msalConfig);
+
 const queryClient = new QueryClient({
   queryCache: new QueryCache({
     onError: (error, query) => {
@@ -198,10 +200,15 @@ export function App() {
       (event: EventMessage) => {
         if (event.payload) {
           const payload = event.payload as AuthenticationResult;
-          setAccessToken(payload.accessToken);
-          patchAccessTokenMutate({
-            body: { id: "smda_api", key: payload.accessToken },
-          });
+          if (event.eventType === EventType.LOGIN_SUCCESS) {
+            const account = payload.account;
+            msalInstance.setActiveAccount(account);
+          } else if (event.eventType === EventType.ACQUIRE_TOKEN_SUCCESS) {
+            setAccessToken(payload.accessToken);
+            patchAccessTokenMutate({
+              body: { id: "smda_api", key: payload.accessToken },
+            });
+          }
         }
         return () => {
           if (id !== null) {
@@ -209,7 +216,7 @@ export function App() {
           }
         };
       },
-      [EventType.ACQUIRE_TOKEN_SUCCESS],
+      [EventType.LOGIN_SUCCESS, EventType.ACQUIRE_TOKEN_SUCCESS],
     );
   }, [msalInstance, patchAccessTokenMutate]);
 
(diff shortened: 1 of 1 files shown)


frontend/src/main.tsx
```diff
@@ -174,20 +222,43 @@ export function App() {
...
+
+  msalInstance.addEventCallback(
+    (event: EventMessage) => {
+      if (event.payload) {
+        const payload = event.payload as AuthenticationResult;
+        const account = payload.account;
+        msalInstance.setActiveAccount(account);
+      }
+    },
+    [EventType.LOGIN_SUCCESS],
```
joargr, line 248: Setting the account as the active one here should be ok, as the user has presumably selected the correct account to log in to.

frontend/src/main.tsx
```diff
@@ -125,14 +140,28 @@ const router = createRouter({
...
+    ...sessionCreateSessionMutation(),
     meta: { errorPrefix: "Error creating session" },
   });
+  const { mutate: patchAccessTokenMutate } = useMutation({
+    ...sessionPatchAccessTokenMutation(),
+    onSuccess: () => {
+      void queryClient.invalidateQueries({
+        queryKey: smdaGetHealthQueryKey(),
+      });
+    },
```
joargr, line 160: A successful addition of the access token to the session will result in the invalidation of the SMDA health query. The health query will thus execute on the API again, leading to an updated result. Were the previous health query failing, it should now be successful, and the updated result will show on the SMDA page.

frontend/src/main.tsx
```diff
@@ -159,11 +188,30 @@ export function App() {
...
+        }
+        return () => {
+          if (id !== null) {
+            msalInstance.removeEventCallback(id);
+          }
+        };
+      },
+      [EventType.ACQUIRE_TOKEN_SUCCESS],
+    );
+  }, [msalInstance, patchAccessTokenMutate]);
```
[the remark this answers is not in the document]
joargr: This callback function has been updated in a newer commit to also set the logged in account as the active account, based on the `LOGIN_SUCCESS` event.

frontend/src/main.tsx
joargr: This code is now outdated, and there is no initialization of the MSAL instance outside of the `App` component. This setting of the active account now happens in the callback function defined inside the `App` component.

PR #48, opened 2025-08-21
ENH: Update sidebar and routes

### LINKED ISSUE #47: Rename 'Directory' in sidebar to 'Project'
Rename `Directory` in the sidebar to `Project` to align it more with current terminology.

tnatt:
Resolves #47 

PR with updates to the sidebar and routes.

- Menu item "Directory" is removed
- Menu item "Mappings" is removed - to be added later when we start on the implementation of this
- Menu item "General" is renamed to "Project", and URL changed from /general to /project
  - Submenu item "SMDA" was renamed to "Masterdata", and the URL was changed
  - Submenu item "Overview" now contains the project selection (previously located in the "Directory" page)

## Checklist


tnatt:
> I think we can even do some more renaming and rearranging of the menu. The main menu item "General" is meant for items related to the project settings, where masterdata/SMDA is the one feature we are currently implementing. After "General" and "Directory" were added to the menu, the item "User" was added. "General" thus should be renamed, and I think it makes sense to put the directory selection (actually project selection) under what now is called "General". In other words, there should only be a main menu item "Project", with the masterdata/SMDA subitem underneath that, and with the project selection on the main page of the "Project" item (called "Overview").
> 
> Based on the current menu structure (before the changes in this PR):
> 
> * Menu item "Directory" is removed
> * Menu item "General" is renamed to "Project", and URL changed from `/general` to `/project`
> * Submenu item "Project / Overview" will contain the project selection, currently located in the "Directory" page

Yes, agree that it needs more rearranging than what currently are present now. So if you think we should do it now, I can give it a go to change this PR :) ?

tnatt:
image
Like so @joargr ?

tnatt:
> > image Like so @joargr ? Mappings also will belong to the project, should I place it under there now while at it?
> 
> Yes, this looks good. Regarding Mappings, I think we could remove that. It doesn't have any proper content, and the only reason it was added to the menu was for filling the menu, to have a handful of items there. Once it's time to implement mappings then we can add it, and make a decision on where to place it - as an item under Project or as a main menu item.

Great, I'll update the PR with these changes then 👍

joargr:
Another change that we can do now, since we are doing several renamings, is to rename "SMDA" to "Masterdata". That's the term we've decided to use, so we should update the GUI with that. As the API endpoints still use the term "SMDA", we should keep the term in the code, for function names etc. So only the names in the menu and page text, as well as the URL, should be renamed.

joargr (approved):
This looks good!

PR #50, opened 2025-08-22
Check for and show project

### LINKED ISSUE #49: Check for and show project
Implement checking for an active project. Show basic information on an active project in relevant places, and control access to features based on whether a project is set.

joargr:
Resolves #49 

Checks if there is a project attached to the session.

- If there is an active project:
  - Shows basic information on the Project page
  - Shows the project name in the header
  - Shows available submenu items
- If there is not active project:
  - Displays the error message on the Project page
  - Hides project submenu items
- A new service for getting project status and data is used. This takes the result of the `projectGetProject` API call and returns a new object, which makes it easy to get information on whether a project exists
- Introduces EDS Tokens, for setting colours on various UI elements
- Note that currently the header might show that there is not a project set. This is a known bug and is related to the session creation. This will be fixed in a separate issue
- Implementing selection of a project will be done in a separate issue

## Checklist


frontend/src/components/Sidebar.tsx
```diff
@@ -2,7 +2,21 @@ import { SideBar as EdsSideBar } from "@equinor/eds-core-react";
...
+  label: string;
+  to: string;
+};
+
 export function Sidebar() {
+  const { data: project } = useProject();
+
+  const ProjectSubItems: AccordianSubItem[] = [];
+  if (project.status) {
+    ProjectSubItems.push({ label: "Masterdata", to: "/project/masterdata" });
```
tnatt, line 17: Are we sure we want to hide this menu item completely, compared to disabling it?

joargr: This is a good question. I think disabling this submenu item could be a better option, however, EDS only supports disabling main menu items, not submenu items. Disabling such a submenu item would probably require adding a non-trivial amount of CSS code and maybe also do some scripting. So right now it is easier to simply hide it.

And then there's another issue. Perhaps in the future we will have use cases where certain features are not relevant for a project. Perhaps a feature is used for an oil and gas project, but not relevant for a CCS project. The submenu for a CCS project would then show only the submenu items for the relevant features. Hiding certain submenu items can then also be done when no active project is set. In other words, the user will experience that the menu and submenus are dynamic.

PR #56, opened 2025-09-01
Avoid parallel session creations

### LINKED ISSUE #54: Perform actions when creating API session
When a session to the API is created, either for the first time or when renewing a session, the following actions needs to be performed:

- Do a refetch of the project data. This should mitigate any no-active-session errors
- Add the access_token to the session, if it exists. This will ensure working with SMDA masterdata can be done

joargr:
Resolves #54

Fixes the bug where multiple API requests returning a status of unauthorized access (ie. missing a session to the API) results in multiple session creation requests. Also ensures that if an SSO access token is present it will be added to the newly created session. Also ensures that an unauthorized access status from getting the project data results in a retry, the same as for other API requests.

## Checklist

### DIFF
--- a/frontend/src/main.tsx
+++ b/frontend/src/main.tsx
@@ -34,6 +34,8 @@ import {
 import { client } from "#client/client.gen";
 import { msalConfig } from "#config";
 import {
+  createSessionAsync,
+  handleAddSsoAccessToken,
   isApiTokenNonEmpty,
   queryAndMutationRetry,
   responseInterceptorFulfilled,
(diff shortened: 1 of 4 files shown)


frontend/src/main.tsx
```diff
@@ -176,18 +181,46 @@ export function App() {
...
         client.instance.interceptors.response.eject(id);
+        setHasResponseInterceptor(false);
       }
     };
   }, [createSessionMutateAsync, apiToken, apiTokenStatus.valid]);
 
+  useEffect(() => {
+    async function callCreateSessionAsync() {
+      await createSessionAsync(createSessionMutateAsync, apiToken);
+    }
```
joargr, line 200: When calling an async function in a `useEffect` callback function it's necessary to define a function inside the callback function, and then call this. The reason is that the callback function itself is a synchronous function, and cannot return a promise (which is the case when calling an async function). Note the `void` when calling this internal function, signaling that there's no promise that should be resolved.

frontend/src/main.tsx
```diff
@@ -176,18 +181,46 @@ export function App() {
...
+
+    if (requestSessionCreation) {
+      if (
+        sessionCreatedAt === undefined ||
+        Date.now() - sessionCreatedAt > 10_000
```
joargr, line 205: At least 10 s have to pass before the application attempts to create a session again. A session usually lasts an hour, so there should not be a situation where such a short waiting period would be problem.

PR #61, opened 2025-09-02
ENH: Add project select dialog

### LINKED ISSUE #18: Set project directory
Implement setting a project directory, using the necessary API endpoint.

tnatt:
Resolves #18 

PR to implement the option to select a project.
A Select project button is introduced in the project page. Clicking this will open a dialog where the user can select from a list of recent project directories or enter a path to a project manually.

- If a recent project directory is clicked, the project is set and the dialog is closed.

- If a manual path is entered  two things can happen
  - it is a valid path, the project is set and the dialog is closed.
  - it is an invalid path, a new dialog with option to initialize a new project is given to the user. 

A helper text is used to display errors to the user in addition to the regular toast. 
If the input for a manual path has any value, the  recent project directory select is disabled.

Outstanding:

- [x]  we lack the option to distinguish between a project folder not existing and the .fmu is not present.  It is only in the latter case that the option to initialize a new project should be given to the user. This needs to be implemented in the API first.
- [x] we need to catch the error coming from the API if a .fmu is not found. This should only trigger the dialog to initialize a new project. 
- [ ] we should implement in the API that user recent_project_directories is updated if a path no longer exists. When this is in place we can handle it through the GUI and refetch recent_project_directories.

  
image
image

## Checklist


frontend/src/components/project/overview/projectselector.tsx
```diff
@@ -0,0 +1,310 @@
...
+            }}
+          />
+        </form.AppForm>
+      </form>
+    </ProjectSelectorFormContainer>
+  );
+}
+function CancelButton({ onClick }: { onClick: () => void }) {
+  return (
+    <Button variant="ghost" onClick={onClick}>
```
joargr, line 175: The cancel button should be [medium-emphasis](https://eds.equinor.com/0b0c666ab/v/0/p/83185a-button/t/e9bf1610925d1aab34c0), and thus be of the `outlined` variant. See also the editing of the user subscription keys.

frontend/src/components/project/overview/ProjectSelector.tsx
```diff
@@ -0,0 +1,311 @@
...
+        open={isDialogOpen}
+        onClose={handleClose}
+        isDismissable={true}
+      >
+        <Dialog.Header>
+          <Dialog.Title>Select project</Dialog.Title>
+        </Dialog.Header>
+        <Dialog.CustomContent>
+          <ProjectSelectorForm closeDialog={handleClose} />
+        </Dialog.CustomContent>
```
joargr, line 307: I think it would be better to have the action buttons (select and cancel) be part of the `Dialog.Actions` element. That probably means that the whole `Dialog` component should be inside the `ProjectSelectorForm` component, but that should be ok.

frontend/src/components/project/overview/ProjectSelector.tsx
```diff
@@ -0,0 +1,311 @@
...
+
+  const queryClient = useQueryClient();
+  const { mutate, isPending } = useMutation({
+    ...projectPostProjectMutation(),
+    onSuccess: () => {
+      void queryClient.invalidateQueries({
+        queryKey: projectGetProjectQueryKey(),
+      });
+    },
+  });
```
joargr, line 65: A couple more options should be defined - using the standard retry function and defining a default error message prefix:
```
  retry: (failureCount: number, error: Error) =>
    queryAndMutationRetry(failureCount, error),
  meta: { errorPrefix: "Error selecting project" },
```

frontend/src/components/project/overview/ProjectSelector.tsx
```diff
@@ -0,0 +1,332 @@
...
+              </form.AppField>
+              <PageSectionSpacer />
+              <form.AppField name="projectPath">
+                {(field) => (
+                  <field.TextField
+                    label="Or enter path to project"
+                    setSubmitDisabled={setSubmitDisabled}
+                  />
+                )}
+              </form.AppField>
```
joargr, line 169: Suggest that the value be validated. As it should be an absolute path, it has to start with "/". It can also be logical to add a maximum length of the path, perhaps 256 characters. When the application knows that the typed value is not value, a call to the API is saved.

The application uses Zod for validation, and there are a couple of general validators defined in `util/validator.ts`. In this case there are two rules that should be combined, so it's easier to just specify the rules directly on the field. The following can be added as a property on `form.AppField`:

```
  validators={{
    onBlur: z.string().max(256).startsWith("/"),
  }}
```

frontend/src/components/project/overview/ProjectSelector.tsx
```diff
@@ -0,0 +1,332 @@
...
+                isOpen={initConfirmDialogOpen}
+                closeDialog={() => {
+                  setInitConfirmDialogOpen(false);
+                }}
+              />
+            </form.AppForm>
+          </ProjectSelectorFormContainer>
+        </Dialog.CustomContent>
+        <form.AppForm>
+          <Dialog.Actions>
```
joargr, line 183: I think `form.AppForm` should be placed inside `Dialog.Actions`. Then the three `Dialog` components will be siblings, and the connection between `form.AppForm` and its components (the buttons) will be clearer.

frontend/src/components/project/overview/ProjectSelector.tsx
```diff
@@ -0,0 +1,332 @@
...
+      <NativeSelect
+        label="Select from recent project directories"
+        id="recent-project-directories"
+        size={5}
+        disabled={disabledSelect}
+        value={disabledSelect ? [] : [field.state.value]}
+        onChange={(e: ChangeEvent<HTMLSelectElement>) => {
+          field.handleChange(e.target.value);
+        }}
+        multiple={true}
```
joargr, line 233: It should not be possible to select multiple projects, so this property should not be specified.

tnatt: I think setting it to false converts it into a dropdown.

frontend/src/components/project/overview/ProjectSelector.tsx
```diff
@@ -0,0 +1,332 @@
...
+
+  return (
+    <InputWrapper
+      helperProps={{
+        text: "",
+      }}
+      helperIcon={<Icon data={error_filled} size={18} />}
+    >
+      <NativeSelect
+        label="Select from recent project directories"
```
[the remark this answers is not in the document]
tnatt: Agree 🙂

frontend/src/components/project/overview/ProjectSelector.tsx
joargr: OK, then it needs to be specified like this. And it wouldn't actually be possible for the user to select multiple options, as the form submits on the first selection.

frontend/src/components/project/overview/ProjectSelector.tsx
```diff
@@ -0,0 +1,332 @@
...
+                    }
+                    recentDirectories={userData.recent_project_directories}
+                  />
+                )}
+              </form.AppField>
+              <PageSectionSpacer />
+              <form.AppField name="projectPath">
+                {(field) => (
+                  <field.TextField
+                    label="Or enter path to project"
```
joargr, line 165: I think the a better wording for this field is the following: "Alternatively, enter a path to the project". This wording makes it clearer that specifying a path is the alternative to selecting a path from a list, and can be used when the wanted path is not present in the list.

frontend/src/components/project/overview/ProjectSelector.tsx
```diff
@@ -0,0 +1,332 @@
...
+        <Dialog.Header>
+          <Dialog.Title>Select project</Dialog.Title>
+        </Dialog.Header>
+        <Dialog.CustomContent>
+          <ProjectSelectorFormContainer>
+            <InputWrapper
+              color="error"
+              helperProps={{
+                text: helperText,
+                icon: <Icon data={error_filled} size={18} />,
```
joargr, line 140: I think the `icon` property can be added, it doesn't seem to have an effect. The error message will be shown with an icon also without this property.

frontend/src/components/project/overview/ProjectSelector.tsx
joargr: I see that there's a problem with the suggested code. It should be possible for an empty value of the field (`projectPath`), as the user can select a path from the recent projects list (`recentProjectPath`). The validator needs to allow for an empty string. Further, it could be useful to define such a validator in the `utils/validator.ts` file, so that it can be used other places.

The following function should work as described:

```ts
export function validatorOptionalMaxLengthStartsWith(
  maxLength: number,
  startsWith: string,
) {
  return z
    .string()
    .refine(
      (val) =>
        val === "" || (val.length <= maxLength && val.startsWith(startsWith)),
      {
        error: `Value must be empty, or start with "${startsWith}" and be a maximum of ${String(maxLength)} characters long`,
      },
    );
}
```

The function can later be refined or remade into a more generic function, if there is a need to vary or combine different rules (ie. not optional, minimum length instead of maximum length, etc.).

frontend/src/components/project/overview/ProjectSelector.tsx
joargr: I see that it is needed when showing an error message that comes from the API, so then it needs to be present.

frontend/src/components/project/overview/ProjectSelector.tsx
```diff
@@ -0,0 +1,332 @@
...
+  return (
+    <Button variant="outlined" color="secondary" type="reset" onClick={onClick}>
+      Cancel
+    </Button>
+  );
+}
+function RecentProjectSelectField({
+  disabledSelect,
+  recentDirectories,
+}: {
```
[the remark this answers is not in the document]
tnatt: should be more consistent now

frontend/src/components/project/overview/ProjectSelector.tsx
```diff
@@ -0,0 +1,332 @@
...
+                listeners={{
+                  onChange: () => {
+                    void form.handleSubmit();
+                  },
+                }}
+              >
+                {(field) => (
+                  <field.RecentProjectSelectField
+                    disabledSelect={
+                      !!form.state.values.projectPath ||
```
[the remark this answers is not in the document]
tnatt: functionality has been changed

frontend/src/components/project/overview/ProjectSelector.tsx
tnatt: ..and now I see your second response 😄

frontend/src/components/project/overview/ProjectSelector.tsx
```diff
@@ -0,0 +1,331 @@
...
+    </ExpansiveDialog>
+  );
+}
+function CancelButton({ onClick }: { onClick: () => void }) {
+  return (
+    <Button variant="outlined" color="secondary" type="reset" onClick={onClick}>
+      Cancel
+    </Button>
+  );
+}
```
joargr, line 209: There should be an empty line before and after such functions.

tnatt: are there no way in typescript to get some autoimatic formatting rules in place? This will take some time to learn 🙃

frontend/src/components/project/overview/ProjectSelector.tsx
```diff
@@ -0,0 +1,331 @@
...
+                      userData.recent_project_directories.length === 0
+                    }
+                    recentProjects={userData.recent_project_directories}
+                  />
+                )}
+              </form.AppField>
+
+              <PageSectionSpacer />
+
+              <form.AppField name="projectPath">
```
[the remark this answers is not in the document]
joargr: Yes, something like that. But I think it would be more natural if the helper text clears `onBlur`, not `onChange`. For instance, if the error message is because the user had a typo in the path, it would be useful to see the original path with typo (shown in the error message) when correcting the path in the field.

frontend/src/components/project/overview/ProjectSelector.tsx
```diff
@@ -0,0 +1,396 @@
...
+                />
+              )}
+            </form.AppField>
+
+            <PageSectionSpacer />
+
+            <form.AppField
+              name="projectPath"
+              validators={{
+                onSubmit: validatorAbsolutePath(),
```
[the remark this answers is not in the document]
tnatt: I struggled with having it onBlur.. the error message was just hanging there and I did not get the submit button to be non disabled easily.
Have you tried?

frontend/src/components/project/overview/ProjectSelector.tsx
```diff
@@ -0,0 +1,396 @@
...
+        },
+        onError: () => {
+          closeDialog();
+        },
+      },
+    );
+  };
+
+  return (
+    <ExpansiveDialog open={isOpen} isDismissable={false} onClose={closeDialog}>
```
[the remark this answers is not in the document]
tnatt: Hm but would you not then close the ProjectSelector dialog everytime the Initialize project dialog is closed by the cancel button ?

frontend/src/components/project/overview/ProjectSelector.tsx
```diff
@@ -0,0 +1,396 @@
...
+  const [valueSource, setValueSource] = useState<ValueSource>("");
+  const codes = [403, 404, 409];
+
+  const closeProjectSelector = ({ formReset }: { formReset: () => void }) => {
+    closeDialog();
+    sethelperTextRecentProjects("");
+    setHelperTextProjectPath("");
+    setValueSource("");
+    formReset();
+  };
```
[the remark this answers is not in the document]
tnatt: done

joargr:
> I see a problem when doing the following:
> 
> * Enter an invalid value in `projectPath`
> * An error message for this field is shown when it is validated (currently on submit)
> * Selecting a path from the `recentProjectPath`
> * Nothing happens. I was expecting that the selected path would be chosen as the project (API call performed), and the dialog would close
> 
> It seems that the normal selection process (`onChange` handler?) doesn't work when there is a validation error message on the `projectPath` field. If there is an error message from an API call the selection from `recentProjectPath` works as expected. And if there is a validation error message on the `projectPath` field entering a new, valid value will trigger the API call as expected.

Now that there is no validation on the fields then this error doesn't occur, so the expected behaviour does happen.

frontend/src/components/project/overview/ProjectSelector.tsx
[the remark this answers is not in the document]
joargr: Yes, sorry, the comment is for the wrong `ExpansiveDialog` component. I meant to refer to the `ExpansiveDialog` component in `ProjectSelectorForm`, not `ConfirmInitProjectDialog`. I think calling `closeProjectSelector` in `onClose` there should be ok.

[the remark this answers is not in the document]
joargr: I tested, and it seems that the `onClose` handler only runs when the dialog is closed through dismissal, ie. by clicking outside the dialog or pressing the Esc key. Closing the dialog from the `Select` or `Cancel` buttons does not trigger the `onClose` handler. Since `isDismissable` is set to `false` it means that the `onClose` handler will never run. Then it would be best to just remove it. The `ConfirmInitProjectDialog` component is also not dismissable, so the same can be done there.

frontend/src/components/project/overview/ProjectSelector.tsx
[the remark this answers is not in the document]
tnatt: Validators skipped in this PR as commented below

PR #65, opened 2025-09-08
ENH: Add option to prevent default error handling

### LINKED ISSUE #63: Allow preventing the default error handling for queries and mutations
ATM it is not possible to override the default error handling from individual query and mutation functions, so a toast and a console error will always be given on errors. 

We should move the default error handling for queries and mutations from the `queryCache` and `mutationCache` to `defaultOptions`.
This error handling should be moved to a separate function in utils e.g  `defaultErrorHandlingView`

This will allow us to override the default behaviour by specifying an `onError` function on the individual query and mutation functions.

tnatt:
Resolves #63 

PR to add a new property in `meta` that can be used to prevent the default error handling.

## Checklist

### DIFF
--- a/frontend/src/utils/query.ts
+++ b/frontend/src/utils/query.ts
@@ -1,0 +1,15 @@
+import { isAxiosError } from "axios";
+import { toast } from "react-toastify";
+
+export const defaultErrorHandling = (error: Error, errorPrefix: string) => {
+  const message =
+    `${errorPrefix}: ` +
+    (isAxiosError(error) &&
+    error.response?.data &&
+    "detail" in error.response.data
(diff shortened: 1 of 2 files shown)


frontend/src/utils/query.ts
```diff
@@ -0,0 +1,15 @@
+import { isAxiosError } from "axios";
+import { toast } from "react-toastify";
+
+export const defaultErrorHandling = (error: Error, errorPrefix: string) => {
+  const message =
+    `${String(errorPrefix)}: ` +
```
joargr, line 6: Using `String()` here shouldn't now be necessary, as the parameter type is set to `string`?

PR #81, opened 2025-09-25
STY: Add stylistic for newline padding

tnatt:
Resolves #issue

Briefly describe changes here, e.g. "Added a precise description about the
changes made to the code, and what the PR is solving."

## Checklist


frontend/eslint.config.js
```diff
@@ -33,10 +34,16 @@ export default tseslint.config(
...
       "react-x": reactX,
+      "stylistic": stylistic,
     },
     rules: {
       ...reactDom.configs.recommended.rules,
       ...reactHooks.configs.recommended.rules,
+      "stylistic/padding-line-between-statements" : [
+        "error", 
+        { blankLine: "always", prev: "*", next: "function" },  
+        { blankLine: "always", prev: "block", next: "return" }
```
joargr, line 45: This subrule is probably too limited, as it only covers code where the previous statement is a block. I think it would be ok to change that to `*`, so that there should be a blank line before a `return` whatever the previous statement is - unless there is nothing else prior (ie. a function that only contains a `return`).

tnatt: Its a bit invasive then.. hence I did not add the `*`. But I'll add a new commit for you to judge

PR #91, opened 2025-10-17
ENH: Display lock information and disable editing

### LINKED ISSUE #88: Indicate whether a project has been opened in read-only mode
The API now knows if a project has been opened in read-only mode, and can return this

https://github.com/equinor/fmu-settings-api/blob/6d6724891f2a08c0c9bcdab40d719d8f5ac5d3fd/src/fmu_settings_api/models/project.py#L28-L29

Read-only mode indicates that two people have opened an FMU Project with FMU Settings. This is indicated by the presence of a lock file that is temporarily placed in the .fmu directory when FMU Settings opens it:

https://github.com/equinor/fmu-settings/blob/main/src/fmu/settings/models/lock_info.py

The information contained in this lock file can now be retrieved from an API route, as well:

https://github.com/equinor/fmu-settings-api/blob/6d6724891f2a08c0c9bcdab40d719d8f5ac5d3fd/src/fmu_settings_api/v1/routes/project.py#L443-L460

This means the GUI can also display helpful information to the user about whom the lock is currently held by, what TGX machine they are on, etc. It also means that the GUI can theoretically query the status of this lock file from time to time. The route may return the lock file currently held by the user, **not** in read-only mode.

An important consequence of this is that attempting to _save_ data, data that has been modified and sent to the api for a write, will now return an error condition if the project is currently in read-only mode. It may still do this if it's not in read-only mode, but the `fmu.settings` has detected it no longer holds the lock (i.e. the project is **not** in read-only mode, but someone has done something like manually deleted the lock file and hijacked writability)

https://github.com/equinor/fmu-settings-api/blob/6d6724891f2a08c0c9bcdab40d719d8f5ac5d3fd/src/fmu_settings_api/v1/routes/project.py#L529-L533

Subissues:
- Display when a project is in read-only mode
- Disallow modifying editable fields if so
- Catch if a patch save returns a permission error, still
- Potentially query the lock file periodically

tnatt:
Resolves #88 

PR  which adds a lock status to the project overview page. 
Lock status is queried on a regular time interval

The read only state is found using the `FmuProject.is_read_only` property.
There are currently three modes the session can be in:
1. `is_read_only=False` - the project is locked by the current session
3. `is_read_only=True` and **lock info is available** - the project is locked by another session
4. `is_read_only=True` and **lock info is undefined** - the project is no longer locked by another session

| mode 1 | mode 2 | mode 3 |
| -- | --|--|
|image | image| image|

For mode 2, the lock information can be displayed
 
When going from mode 2 to mode 3 a toast will appear informing the user that the project can be enabled for editing, and a button for acquiring the lock appears. Pressing this will lock the project and set the project read only state to `False.`

## Checklist


frontend/src/components/LockStatus.tsx
```diff
@@ -0,0 +1,177 @@
...
+    <>
+      <Button
+        variant="outlined"
+        color="primary"
+        ref={anchorRef}
+        onClick={openPopover}
+        aria-haspopup
+        aria-expanded={isOpen}
+      >
+        Lock information
```
joargr: The button text should be a verb, so the text should be "Show lock information".

frontend/src/components/LockStatus.tsx
```diff
@@ -0,0 +1,177 @@
...
+}
+
+export function LockIcon({ isReadOnly }: { isReadOnly: boolean }) {
+  return (
+    <Icon
+      data={isReadOnly ? lock : lock_open}
+      color={
+        isReadOnly
+          ? tokens.colors.infographic.primary__energy_red_100.rgba
+          : tokens.colors.infographic.primary__moss_green_100.rgba
```
joargr: I think this might be the wrong colours from the EDS palette. The infographic colours are meant for diagrams, whereas icon colours should come from the `interactive` set. The EDS documentation has just been updated just now, so the previous colour list seems to no longer be available, but the new [colour documentation](https://eds.equinor.com/docs/foundation/colour/intro) says "colours in the _Text_ group are designed specifically for text and icons, ensuring the right contrast and readability.".

The application is still using the old format, defining colours using tokens (as opposed to the new format using CSS variables), so the most likely colours for the the two states of the lock icon should be these:

- read-only: `tokens.colors.interactive.success__resting.hex`
- editable: `tokens.colors.interactive.danger__resting.hex`

The actual value accessed should be the `hex` value, as this is also used elsewhere in the application.

frontend/src/components/LockStatus.tsx
```diff
@@ -0,0 +1,177 @@
...
+  }, [lockStatus?.lock_info, isReadOnly, queryClient]);
+
+  return (
+    <Banner>
+      <Banner.Icon>
+        <LockIcon isReadOnly={isReadOnly} />
+      </Banner.Icon>
+
+      <Banner.Message>
+        {`Project is ${isReadOnly ? "read only" : "editable"}`}
```
joargr: The term is usually spelled "[read-only](https://en.wikipedia.org/wiki/Read-only)", with the hyphen, so this spelling should be used consistently throughout the application. Variables etc. in the application are mostly spelled using two words (ie. `isReadOnly`), and don't have to be renamed.

frontend/src/components/LockStatus.tsx
```diff
@@ -0,0 +1,177 @@
...
+          <Table.Cell>{lock_info.hostname}</Table.Cell>
+        </Table.Row>
+        <Table.Row>
+          <Table.Cell>Locked since</Table.Cell>
+          <Table.Cell>{displayTimestamp(lock_info.acquired_at)}</Table.Cell>
+        </Table.Row>
+        <Table.Row>
+          <Table.Cell>PID</Table.Cell>
+          <Table.Cell>{lock_info.pid}</Table.Cell>
+        </Table.Row>
```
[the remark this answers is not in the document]
tnatt: ..hehe no most likely not, I am just traumatized by processes hanging in the background 😄. I'll remove it

frontend/src/components/LockStatus.tsx
joargr: I agree that the green in `success__resting` is quite green. Using a black colour seems better, although even this colour should be specified using a token (maybe "black" in EDS isn't actually the regular `#000000`). However, I see that the colour token `primary__resting` has a relevant name and it is indeed green, so using this colour token for editable mode seems quite appropriate.

PR #113, opened 2025-10-31
ENH: Rename User to User settings

### LINKED ISSUE #79: Rename `User` menu to `User Settings`
In the accordion menu we have a `User` option to display user settings. It might be more clear what this menu is for if we rename it to `User Settings`.

This change should also be reflected in component names, directories, and routes, to keep the project understandable consistent during development.

tnatt:
Resolves #79 

PR which renames `User` in the sidebar to `User settings` and updates the route name accordingly,

## Checklist


joargr:
Hmm, I'm not fully onboard on renaming this menu item to "User settings". While it might make it clearer to the user that this menu item is about "user settings" - managing settings related to the user - I can see a couple of arguments against this renaming:

1. Why is only "User" renamed, and not "Project" as well? The same argument for having "User settings" can also be used for having "Project settings"
2. The application is called "FMU Settings", so it's implied that both "Project" and "User" have something to do with "settings". Thus it should not be necessary to be explicit about it in the menu
3. If the menu item is renamed "User settings", does that limit what can be put under that main menu item? What if there's a need in the future to put something user related in the menu that is not related to settings. An example might be "tasks", something that the user must act on; or displaying a changelog, showing all user activity and not just settings changes

I think it would be good to keep the menu item label as "User", as it would not limit what can be put there as sub-items in the future. But perhaps we can have a discussion on this before proceeding with this PR.

PR #117, opened 2025-11-03
Masterdata editing

### LINKED ISSUE #44: Implement editing of masterdata
Implement editing of masterdata. The data will come from two sources:
- The `smdaPostMasterdataOptions` API call. Given a list of one or more fields (assets), it will return masterdata from the SMDA API
- The masterdata in the project object, from the `projectGetProject` API call. If there is no existing project masterdata, this will be empty

The data from the two sources should be reconciled, and any differences should be indicated. For instance, if there is a discovery value that is not present in the existing project masterdata, this discovery value could be indicated so that the user can add it to the project. Similarly, if a discovery value is in the project masterdata but not in the SMDA masterdata, this should be indicated so that the user can remove it from the project masterdata (perhaps in this case the storing of the project masterdata will fail in the API because of this).

This feature will be used both for adding new project masterdata (when no project masterdata exists) and for editing existing project masterdata.

joargr:
Resolves #44 

Allows for editing of project masterdata. Contents of existing project masterdata is shown on the left, along with any additional masterdata from SMDA for the existing fields (called "available masterdata") shown on the right (see picture 1). Gives the possibility to search for additional fields (see picture 2), which will be added to the available masterdata.

- Fields, countries and discoveries can be added from the available masterdata to the project masterdata by clicking on the chip (it has a left arrow), and can be removed from the project masterdata by clicking on the chip (where it has a right arrow)
- The options for stratigraphic column are taken from all fields in the project and in the available masterdata
- The options for coordinate system are always the same for all fields
- Discoveries are grouped on the fields they belong to
- If an available discovery belongs to a field that is not part of the project masterdata, the discovery can not be added, and the chip does not have a left arrow
- If the project contains a discovery belonging to a field that is not part of the project, the discovery is marked as being removed from the project when the masterdata is saved (see picture 3)
- Includes a general function `httpValidationErrorToString` for presenting a validation error from an API response as a string, formatting it based on the fields referenced in the Pydantic error message
- Adds some types and functions for processing objects as defined in the API models, notably a "name-uuid" object, containing an `identifier` and `uuid` pair, which the objects that are part of the masterdata have

Picture 1

Picture 2

Picture 3

While this PR contains the full functionality to display and edit project masterdata, there are still some features and small bugs left to resolve. These will be implemented as separate issues.

## Checklist

### DIFF
--- a/frontend/src/components/project/masterdata/Edit.tsx
+++ b/frontend/src/components/project/masterdata/Edit.tsx
@@ -513,7 +513,7 @@ export function Edit({
         closeDialog={closeSearchDialog}
       />
 
-      <EditDialog open={isOpen}>
+      <EditDialog open={isOpen} $maxWidth="200em">
         <form
           onSubmit={(e) => {
             e.preventDefault();
--- a/frontend/src/components/project/masterdata/FieldSearch.tsx
+++ b/frontend/src/components/project/masterdata/FieldSearch.tsx
@@ -122,7 +122,7 @@ export function FieldSearch({
   };
 
   return (
-    <EditDialog open={isOpen}>
+    <EditDialog open={isOpen} $maxWidth="200em">
       <Dialog.Header>Field search</Dialog.Header>
 
       <Dialog.CustomContent>


GibranAlfa (approved):
looks good to me based on the demo in the standup today 👍

tnatt (approved):
Tested locally and does the job, nice work 🙂 
I have a couple of minor comments that I will gather up into an issue instead, so that we can get this one merged and demonstrated!

PR #126, opened 2025-11-07
BUG: Add the empty option to masterdata initial value check

### LINKED ISSUE #125: Initial select option error message when editing empty masterdata
When editing empty masterdata the option lists of the coordinate system and stratigraphic column form fields are empty, except for the "(none)" entries. The form shows an error saying that the "(none)" option does not exist. There is however such an option. This needs to be fixed.

joargr:
Resolves #125 

Fixes the bug by adding the empty option to the check of initial value in the select options list. Also renames some types for better consistency.

## Checklist

### DIFF
--- a/frontend/src/components/Sidebar.tsx
+++ b/frontend/src/components/Sidebar.tsx
@@ -1,6 +1,6 @@
 import { SideBar as EdsSideBar } from "@equinor/eds-core-react";
 import { account_circle, dashboard, folder } from "@equinor/eds-icons";
-import { Link } from "@tanstack/react-router";
+import { Link, useLocation } from "@tanstack/react-router";
 
 import { useProject } from "#services/project";
 
@@ -11,6 +11,7 @@ type AccordianSubItem = {
 
 export function Sidebar() {
   const project = useProject();
+  const location = useLocation();
 
   const ProjectSubItems: AccordianSubItem[] = [];
   if (project.status) {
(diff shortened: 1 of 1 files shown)


frontend/src/components/project/masterdata/Edit.tsx
```diff
@@ -209,14 +209,14 @@ function createItemLists(
...
+  array: IdentifierUuidType[],
   initialValue: IdentifierUuidType,
 ): void {
   setFieldMeta(field, (meta) => ({
     ...meta,
     errorMap: {
-      onChange: findOptionValueInIdentifierUuidArray(
-        identifierUuidArray,
+      onChange: findOptionValueInNameUuidArray(
+        [emptyIdentifierUuid(), ...array],
```
tnatt: This change here means that an empty value is permitted to submit and the API will respond with a validation error. Is it possible to make it more clear to the user that this actually is a required field. Maybe by splitting the validation for an empty string and the other options?

joargr: Perhaps validation should be a separate topic for discussion. Should there be some validation in the GUI, or is it best to only let the API to do that? And what form fields should have validation - is it ok to save the masterdata with no fields (assets), or is it required to have at least one field?

tnatt: Yes agree, we should align on one way of doing it. 
It was actually a bit puzzling to me that it is allowed to save the masterdata without many of these items, I thought all were required. And my guess is that it is not the intended behavior, and we should revisit the [SMDA model]( https://github.com/equinor/fmu-datamodels/blob/003dc98038a4e2fe72f1094d6a5e34bcda05a107/src/fmu/datamodels/fmu_results/fields.py#L344) and add a length requirement to all fields of list type.

tnatt (approved):
This looks good 👍 
We should however decide how validation on these input fields happen (GUI validation or API response), and look into if the model used to validate the data needs an update. I'll make an issue in `fmu-datamodels`

Issue #132, opened 2025-11-14
Don't use default maxWidth on edit dialog


joargr:
The styling of an edit dialog component is defined in `styles/common.ts`, in `EditDialog`. Attributes to styled-components is used for the `minWidth` and `maxWidth` CSS properties, with default values (`10em` and `20em` respectively). For pages using an edit dialog this allows for setting specific values for these width properties, with default values used otherwise.

While it makes sense to have a default value for the minimum width of an edit dialog, having a default maximum width is problematic. If the page using an edit dialog doesn't explicitly set a value for this, the dialog will get the default value of `20em`. This is quite narrow, and leads to the page having to set an arbitrary high value so as to override the default value.

There shouldn't be defined a default value for maximum width. When deciding the width of the edit dialog, the browser will usually set this based on the content of the dialog. If there is a page with a special need to control the width of the dialog, it can be set in the page itself, when specifying the edit dialog attributes - using `width` or `maxWidth` which will be sent on the EDS `Dialog` component. Thus, specifying attributes as part of the styled component (ie. extra attributes, prefixed with `$`) is mostly used for those cases where there is a need for a default value, and that this value can be overridden.

Update 2025-12-02:
The component is using attributes with default values. The way default values are handled should be updated, to be consistent with how it's done for other components (`PageHeader` and  `PageText`):

```typescript
export const EditDialog = styled(Dialog).attrs<{
  $minWidth?: string;
}>(({ $minWidth = "10em" }) => ({
  style: {
    minWidth: $minWidth,
  },
}))`

...

`;
```

Update 2026-02-18:
A new attribute has been added to the component: `extraPaddingBottom`, with a default value of `true`. The attribute and default value is currently defined in the old way, to be consistent with other attribute definitions, but this should be updated to the new format as part of this issue.

PR #151, opened 2025-12-01
ENH: Add project information on homepage

### LINKED ISSUE #128: Make homepage more project oriented
The homepage now does not reveal anything about the selected project. Since editing an FMU project is the main use of the application this should be emphasized more on the homepage. 

Some thoughts of elements that could be in the homepage

- What project is the selected one
- When it was last edited
- Some simple project health checks to quickly detect what needs to be configured (e.g. masterdata is not added yet)
- Project changelog easily accessible - maybe displaying latest 5 changes with option to open all
-  ...

tnatt:
Resolves #128 

PR which adds some simple information about the opened project to the homepage. 
- path, last modified and model description. 
If the application is opened outside a project, the project selector is displayed.

The information is shown in a grey InfoBox to align with how we display other config contents. This change was also done to the project overview page where the project information was shown on a white background.

image

## Checklist

### DIFF
--- a/frontend/src/components/LockStatus.style.ts
+++ b/frontend/src/components/LockStatus.style.ts
@@ -1,6 +1,17 @@
 import { Banner } from "@equinor/eds-core-react";
+import { tokens } from "@equinor/eds-tokens";
 import styled from "styled-components";
 
 export const LockStatusBanner = styled(Banner)`
   margin-bottom: 1em;
+  border: solid 1px ${tokens.colors.ui.background__medium.hex};
+  border-radius: ${tokens.shape.corners.borderRadius};
+  box-shadow: none;
+
+  /* Adjust elements and ensure clean border corners */
+  [class*=Banner__Content],
+  [class*=Banner__NonMarginDivider] {
+    background: none;
+  }
+
 `;


frontend/src/components/home/Resources.style.ts
```diff
@@ -3,14 +3,12 @@ import { tokens } from "@equinor/eds-tokens";
 import styled from "styled-components";
 
 export const ResourcesContainer = styled.div`
-  display: flex;
-  justify-content: flex-start;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(auto-fit, minmax(250px, 1fr));
```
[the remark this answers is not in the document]
joargr: This is a very nice change!

frontend/src/routes/project/index.tsx
```diff
@@ -9,17 +9,40 @@ import { EditableModelInfo } from "#components/project/overview/Model";
...
+import { Typography } from "@equinor/eds-core-react";
+
+function ProjectInfoBox({ projectData }: { projectData: FmuProject }) {
+  const created_date = displayDateTime(projectData.config.created_at);
+
+  return (
+    <InfoBox>
+      <Typography variant="h3">{projectData.project_dir_name}</Typography>
+
+      <Typography token={{ fontSize: "14px" }}>
```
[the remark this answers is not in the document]
tnatt: Tried to follow Sumo's design here. But that is ok, I can change back

frontend/src/routes/project/index.tsx
```diff
@@ -9,17 +9,40 @@ import { EditableModelInfo } from "#components/project/overview/Model";
...
+
+function ProjectInfoBox({ projectData }: { projectData: FmuProject }) {
+  const created_date = displayDateTime(projectData.config.created_at);
+
+  return (
+    <InfoBox>
+      <Typography variant="h3">{projectData.project_dir_name}</Typography>
+
+      <Typography>
+        {projectData.path}
```
joargr: As the path is an important piece of information to show, I think it should be displayed a bit more prominently. Setting it to bold would make it displayed a bit too prominent, so I suggest showing it with just a little bit more weight. A font weight for normal is 400 whereas the font weight for bold is 700, but for the font used (Equinor) a weight of 500 is available (this is usually referred to as [Medium](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-weight#common_weight_name_mapping)).

I recommend that this be implemented as an inline-level class on `PageText`, called `emphasis`. The project info on this page thus should be shown using this component instead of using the `Typography` component. Since there should not be a bottom margin in this case, the `PageText` should be adjusted to cater for this need. The following update to `styles/common.ts` accomplishes these needs:

```typescript
export const PageHeader = styled(Typography).attrs<{ $variant?: string }>(
  ({ $variant = "h2" }) => ({ variant: $variant }),
)`
  margin-bottom: 0.5em;
`;

export const PageText = styled(Typography).attrs<{
  $variant?: string;
  $marginBottom?: string;
}>(({ $variant = "body_short" }) => ({ variant: $variant }))`
  margin-bottom: ${({ $marginBottom = tokens.spacings.comfortable.medium }) => $marginBottom};

  .emphasis {
    font-weight: 500;
  }
`;
```

This update introduces a `$marginBottom` attribute, where a value for `margin-bottom` can be set when using the `PageText` component. If not specified, the value defaults to `1em` (via token). For the project info on the project overview page, the attribute value can be specified: `<PageText $marginBottom="0">`.

Note that the way default values are dealt with in the component definition is changed. I think this updated way is better and more concise. The same way should be used for `PageHeader` as well, so both components should updated. The component `EditDialog` also uses default values and should be updated, but there's an existing [issue](https://github.com/equinor/fmu-settings-gui/issues/132) on this component, so changing the way default values are defined can be done in that issue.

frontend/src/styles/common.ts
```diff
@@ -2,16 +2,22 @@ import { Chip, Dialog, List, Typography } from "@equinor/eds-core-react";
...
 
-export const PageHeader = styled(Typography).attrs<{ $variant?: string }>(
-  (props) => ({ variant: props.$variant ?? "h2" }),
-)`
-  margin-bottom: 0.5em;
+export const PageHeader = styled(Typography).attrs<{
+  $variant?: string;
+  $marginBottom?: string;
+}>(({ $variant = "h2" }) => ({ variant: $variant }))`
+  margin-bottom: ${({ $marginBottom = tokens.spacings.comfortable.medium }) => $marginBottom};
```
joargr: The `medium` spacing corresponds to `16px` and `1em`, but the default value needs to be `8px` and `0.5em`, which corresponds to the `small` spacing.

frontend/src/components/LockStatus.style.ts
```diff
@@ -1,6 +1,20 @@
 import { Banner } from "@equinor/eds-core-react";
+import { tokens } from "@equinor/eds-tokens";
 import styled from "styled-components";
 
 export const LockStatusBanner = styled(Banner)`
+  border: solid 1px ${tokens.colors.ui.background__medium.hex};
+  border-radius: ${tokens.shape.corners.borderRadius};
+  box-shadow: none;
+
   margin-bottom: 1em;
```
joargr: Suggest moving this to the top, for consistent ordering of the `margin`, `padding` and `border` rules (with no empty lines between `margin-bottom` and `border`).

PR #172, opened 2025-12-10
MAINT: Refactor edit data

### LINKED ISSUE #148: Fixes to select fields
There are some fixes that need to be done to the select form fields (Coordinate system and Stratigraphic column):

- The options lists need to be updated on changes to the form - most significantly, when a field value changes (a field is added to or removed from the project)
- The options should be based on the fields that are part of the project. If a new field is made available (from the field search) then associated values for this field (stratigraphic columns and default coordinate system) should not be updated in the option lists until the field is actually added to the project

joargr:
Resolves #148 

Refactors the handling of edit data. Previously, the data used for the form and for the edit dialog was handled as two different sets of list:  `projectItems`, `availableItems` and `orphanItems` for item lists ("Field", "Country" and "Discovery"), and `optionsData` for select options ("Coordinate system" and "Stratigraphic column"). Now there is one set of lists:

- `projectData`: Contains data for all form fields that are part of the project
- `availableData`: Contains data for all form fields that are not part of the project but available for adding to the project. Only item fields are relevant for this
- `orphanData`: Contains data on orphans - items that belong to a Field but where that Field is not part of the project. Only Discovery orphans are currently supported

The contents of these lists are still handled by separate functions depending on the kind of data processed: `createOptions` and `createItemLists`.

There are also changes to the types (models) used for these edit data lists, with types accurately describing which form fields are relevant for each data list. The types are prefixed `FormMasterdata`, as they describe the structure of these edit data lists, which is different from the masterdata type for the data stored in the project. For each form field, items are grouped on the Field they belong to. Currently only Discovery items use proper Field grouping. For Field and Country items a dummy group is used for grouping, so as to easily manage this data (the item lists for these form fields share the same type).

Other changes in this PR:

- Options for select lists ("Coordinate system" and "Stratigraphic column") are now updated on any change, to any field in the form, similar to how it is for item lists
- Adding a Field from the field search does not affect the select options in the project data
- Some renaming of functions and types

## Checklist


tnatt (approved):
AFAIK this looks good 👍

Issue #174, opened 2025-12-11
Use only lock status to determine project editing status


joargr:
Currently a project's editing status is determined by checking the `isReadOnly` field on the project data. Meanwhile, a project's lock status is fetched regularly (every minute) and gives more specific information on the lock. This data is also used, especially for showing information on the lock itself. When a lock is no longer held, the project data needs to be refreshed so that the `isReadOnly` can have the correct value.

The fact that the GUI uses information from both the project data and the lock status complicates things, as the information from those two sources needs to be in sync. It should be enough to just use information from the lock status, and to disregard the `isReadOnly` field on the project. All places that show information on editing status (read-only mode) should be updated to only use ìs_lock_acquired` on the lock status.

Note that currently a toast is shown when a lock is no longer held. This is possible to detect since there is a moment when `is_lock_acquired` is false and the project's `isReadOnly` is still false. Using only `is_lock_acquired` means that the detection of the status change is not easily possible. However, in the near future there will probably be a feature implemented where the lock expiration data will be used to show a message to the user when a lock expires.

PR #175, opened 2025-12-12
ENH: Masterdata editing disabled when project is read-only

### LINKED ISSUE #173: Ensure no editing when project is in read-only mode
When the project is in read-only mode it should not be possible to edit the masterdata:

- The edit button should not be enabled
- The save button should be disabled

joargr:
Resolves #173 

When the project is read-only the editing of masterdata is disabled:

- The edit button on the masterdata page is disabled
- If the masterdata edit dialog is open, the save button is disabled
- As an extra check, submission of the form data to the API is not done

Determining the read-only status of the project is done by checking `is_lock_acquired` in the lock status data. The `is_read_only` property of the project data is not consulted, as per the decision to be implemented in issue #174. The lock status is checked frequently (every minute), so read-only status might be entered while the edit dialog is open, leading to the save button becoming disabled.

## Checklist


joargr:
> should the `is_read_only` property be removed after #174 is solved? if the gui decide to only fetch lock_status from time to time, then this issue in the api ([equinor/fmu-settings-api#181](https://github.com/equinor/fmu-settings-api/issues/181)) can be closed as well

Yes, I think the GUI does not need the `is_read_only` property, so after #174 is solved the property can be removed. As for where to get the lock status, currently it is a separate endpoint which I think makes sense, but maybe we could have a brief discussion on this in a planning meeting. The alternative would be to have lock status as part of the project, but then the API would have read the whole project frequently (every minute), which might be a problem if the contents of the project data grows over time with new features implemented.

tnatt (approved):
This is good 👍 
I really like the change to using only the `lock_status` to detect the read-only status for the project! I'll do the same for the rest of the GUI code, that will simplify! 🙂

PR #178, opened 2025-12-15
ENH: Add functionality for opening and closing RMS

### LINKED ISSUE #179: Add functionality for opening and closing a RMS project
We should add some action buttons to be able to open and close the selected RMS project.
They should be placed together with the button to select a project.

tnatt:
Resolves #179  

To be able to fetch data from RMS, the RMS project needs to be opened in the session first.
This PR adds buttons to perform these actions - open, reload and closing an RMS project.
To keep track of whether an RMS project has been opened or not the session storage is utilized.

## Checklist

### DIFF
--- a/frontend/src/components/project/rms/Overview.style.ts
+++ b/frontend/src/components/project/rms/Overview.style.ts
@@ -1,0 +1,10 @@
+import { tokens } from "@equinor/eds-tokens";
+import styled from "styled-components";
+
+export const ActionButtonsContainer = styled.div`
+  margin-bottom: ${tokens.spacings.comfortable.medium};
+
+  button + button {
+    margin-left: ${tokens.spacings.comfortable.small};
+  }
+`;
--- a/frontend/src/components/project/rms/Overview.tsx
+++ b/frontend/src/components/project/rms/Overview.tsx
@@ -1,14 +1,16 @@
 import { Dialog } from "@equinor/eds-core-react";
 import { createFormHook } from "@tanstack/react-form";
 import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
-import { useState } from "react";
+import { Dispatch, SetStateAction, useEffect, useState } from "react";
 import { toast } from "react-toastify";
 
-import { FmuProject, RmsProject } from "#client";
+import { RmsProject } from "#client";
 import {
   projectGetProjectQueryKey,
   projectGetRmsProjectsOptions,
   projectPatchRmsMutation,
+  rmsDeleteRmsProjectMutation,
+  rmsPostRmsProjectMutation,
 } from "#client/@tanstack/react-query.gen";
 import {
   CancelButton,
@@ -23,6 +25,12 @@ import {
 } from "#utils/api";
 import { fieldContext, formContext } from "#utils/form";
 import { getRmsProjectName } from "#utils/model";
+import {
+  getStorageItem,
+  STORAGENAME_RMS_PROJECT_OPEN,
+  setStorageItem,
+} from "#utils/storage";
+import { ActionButtonsContainer } from "./Overview.style";
 
 const { useAppForm: useAppFormRmsEditor } = createFormHook({
   fieldComponents: {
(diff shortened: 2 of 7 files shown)


frontend/src/client/sdk.gen.ts
```diff
@@ -229,6 +229,74 @@ export const projectPatchRms = <ThrowOnError extends boolean = false>(options: O
     });
 };
 
+/**
+ * Saves the RMS coordinate system in the project .fmu directory
```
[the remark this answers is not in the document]
tnatt: That's a bit odd.. This file is auto generated from the API using the `pnpm openapi-ts` command, and should not be modified manually. But why it is duplicated I don't really know.

frontend/src/components/project/rms/Overview.tsx
```diff
@@ -209,10 +216,102 @@ function RmsInfo({ rmsData }: { rmsData: RmsProject }) {
...
+  isRmsProjectOpen,
+}: {
+  rmsData: RmsProject | null | undefined;
+  projectReadOnly: boolean | undefined;
+  setIsRmsProjectOpen: React.Dispatch<React.SetStateAction<boolean>>;
+  isRmsProjectOpen: boolean;
+}) {
+  const [selectProjectDialogOpen, setSelectProjectDialogOpen] = useState(false);
+
+  const { mutate: mutateOpen, isPending: isPendingOpen } = useMutation({
```
joargr: An alternative to setting local names for individual functions and variables in a mutation is to set the result of the `useMutation` to a variable itself: `const projectOpenMutation = useMutation(...)`. Then the functions and variables can be referred to as `projectOpenMutation.mutate` and `projectOpenMutation.isPending`. I think this looks better, especially when there is more than one `useMutation`. Then there would be function calls to `projectOpenMutation.mutate` and `projectCloseMutation.mutate`. A component that uses this naming is `frontend/src/components/project/masterdata/Edit.tsx`.

frontend/src/components/project/rms/Overview.tsx
```diff
@@ -209,9 +220,121 @@ function RmsInfo({ rmsData }: { rmsData: RmsProject }) {
...
+}
+
+export function Overview({
+  rmsData,
+  projectReadOnly,
+}: {
+  rmsData: RmsProject | null | undefined;
+  projectReadOnly: boolean;
+}) {
+  const [isRmsProjectOpen, setIsRmsProjectOpen] = useState(() => {
```
joargr: As this state value is not used directly in `Overview`, only passed on to `RmsProjectActions`, I think the definition and `useEffect` should be set in that component.

frontend/src/components/project/rms/Overview.tsx
```diff
@@ -209,9 +220,117 @@ function RmsInfo({ rmsData }: { rmsData: RmsProject }) {
...
+  isRmsProjectOpen,
+}: {
+  rmsData: RmsProject | null | undefined;
+  projectReadOnly: boolean | undefined;
+  setIsRmsProjectOpen: React.Dispatch<React.SetStateAction<boolean>>;
+  isRmsProjectOpen: boolean;
+}) {
+  const [selectProjectDialogOpen, setSelectProjectDialogOpen] = useState(false);
+
+  const projectOpenMutation = useMutation({
```
joargr: If the open project mutation were to fail, the `isRmsProjectOpen` flag should be set to `false`. This could happen if the project has been successfully opened, but then it fails when the project is reloaded.

frontend/src/components/project/rms/Overview.tsx
```diff
@@ -209,9 +220,117 @@ function RmsInfo({ rmsData }: { rmsData: RmsProject }) {
...
-  const rmsData = projectData.config.rms;
+function RmsProjectActions({
+  rmsData,
+  projectReadOnly: projectIsReadOnly,
+  setIsRmsProjectOpen,
+  isRmsProjectOpen,
+}: {
+  rmsData: RmsProject | null | undefined;
+  projectReadOnly: boolean | undefined;
+  setIsRmsProjectOpen: React.Dispatch<React.SetStateAction<boolean>>;
```
joargr: Suggest importing these from `React`, as that's how it is done in other places.

```suggestion
  setIsRmsProjectOpen: Dispatch<SetStateAction<boolean>>;
```

This suggestion also applies to `RmsEditorForm`.

frontend/src/components/project/rms/Overview.style.ts
```diff
@@ -0,0 +1,14 @@
+import { tokens } from "@equinor/eds-tokens";
+import styled from "styled-components";
+
+export const ActionButtonsContainer = styled.div`
+  margin-bottom: ${tokens.spacings.comfortable.medium};
+
+  button {
+    min-width: 9em;
```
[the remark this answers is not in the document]
tnatt: That will work yes 👍 I will construct an issue for the submit button

frontend/src/components/project/rms/Overview.tsx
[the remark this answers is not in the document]
tnatt: I don't think anything happens in the API regarding this, but that would be beneficial if there was a cleanup that closed the project. We can take it up in the standup.
In the GUI if a user selects a new RMS project, at least the state value of `isRmsProjectOpen` is set to `false` on a successful mutation, so the user will see "Open project" instead of "Reload project". And will see the text that indicates that RMS project needs to be opened to access data.

joargr (approved):
This looks very good!

PR #185, opened 2025-12-23
ENH: Introduce warning box component

### LINKED ISSUE #184: Introduce warning box
Today there is a common style component `InfoBox`, for showing information in a box. There should be a similar box for showing a warning. This can be called `WarningBox`, and have a similar look but with a different background colour. As both boxes are quite similar, a generic box can be defined, and both boxes then be based on this generic box.

With these boxes in place, other boxes can be based on `InfoBox` or `WarningBox`, with necessary adjustments added. An example of this would be `OrphanTypesContainer` in the masterdata editing.

joargr:
Resolves #184 

Introduces a styled component for a warning box, similar to the existing info box. Both boxes are based on a generic box, which defines margin, padding and border, along with a basic background colour and content colour. Boxes based on this generic box can override these colours when necessary.

Also, boxes used in the masterdata editing are adjusted so that they use the info and warning boxes as a basis.

The application header uses a `ProjectInfoContainer` component, which has a very similar look to an info box. However, this box has specific definitions for height and padding, so it does not seem like a good candidate to be based on the common info box.

## Checklist

### DIFF
--- a/src/fmu_settings_gui/__main__.py
+++ b/src/fmu_settings_gui/__main__.py
@@ -1,7 +1,6 @@
 """The main entry point for fmu-settings-gui."""
 
 import asyncio
-import os
 import re
 import signal
 import sys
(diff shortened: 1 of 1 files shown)


tnatt (approved):
Nice addition 👍

PR #187, opened 2026-01-05
BUG: Uniform icon size

### LINKED ISSUE #129: Use uniform size for error icon
Error messages for form fields are shown with an error icon (`error_filled`). Currently there are two different sizes specified where this icon is used - 16 and 18. This should be corrected, so that the same size is used consistently. The icons is used along with an error text, as a helper text in the EDS fields, so 16 is probably the preferred size.

Components with the icon size set to 18:
- `frontend/src/components/project/overview/ProjectSelector.tsx`: `ProjectSelectorForm` and `RecentProjectSelect`

The`Select` component, as used in the masterdata editing, also uses the error icon, but setting the correct size of this is handled in a separate [issue](https://github.com/equinor/fmu-settings-gui/issues/127).

Note that in the header there is an icon (`comment`) used for the Feedback link. This also has a size set to 18, so it can be considered if this also should be changed.

nourinmohd:
Resolves #129 

Change the icon size to 16. 

## Checklist

### DIFF
--- a/frontend/src/components/Header.tsx
+++ b/frontend/src/components/Header.tsx
@@ -103,7 +103,7 @@ function FeedbackDialog() {
   return (
     <>
       <HeaderActionButton onClick={openDialog}>
-        Feedback <Icon data={comment} size={16} />
+        Feedback <Icon data={comment} size={18} />
       </HeaderActionButton>
 
       <EditDialog


tnatt (approved):
LGTM

PR #188, opened 2026-01-05
BUG: Use eds space tokens

### LINKED ISSUE #55: Use EDS tokens for spacing
Currently most margin and padding values use the `em` unit for spacing. This should be replaced to use EDS tokens, as the EDS design uses `px` units, in `8px` increments. A list of available spacing tokens is shown in the [documentation](https://eds.equinor.com/0b0c666ab/p/721e1f-spacing).

As the font size used in the application is `16px` (the default font size in the browser), a value of `1em` equals `16px` and thus the EDS spacing token of `Medium`. All margin and padding values should be changed to use this (ie. `tokens.spacings.comfortable.medium`). Other values used are `0.5em` and `2em`, which have corresponding spacing tokens.

A few notes:

- In `frontend/src/styles/common.ts`, the component `PageHeader` has a `margin-bottom` value of `0.5em`. This component will be shown as a HTML header element, which have some special handling when it comes to margins. Therefore, the spacing token to use for this should be `Medium Small` to have (roughly) the same margin as it does currently
- In `frontend/src/styles/common.ts`, the component `PageCode` has a `margin` value of `0 1em 1em 1em` (setting values separately for top, right, bottom and left margins). To avoid having to repeat the spacing token three times, this declaration can be replaced with the two declarations `margin: (token); margin-top: 0;`, where the second declaration for `margin-top` overrides the first declaration for the top margin
- In `frontend/src/styles/common.ts`, the component `PageSectionSpacer` has a declaration for `height`. This component is used for spacing, so in this case the height value can be replaced with a spacing token
- Margin and padding values of 0 will not need to be replaced, in fact there is no EDS spacing token for this value
- Width and height is also used to define the size of areas, and this is not relevant for using spacing tokens

nourinmohd:
Resolves #55

Added spacing tokens to the relevant files. 

## Checklist

### DIFF
--- a/frontend/src/components/LockStatus.style.ts
+++ b/frontend/src/components/LockStatus.style.ts
@@ -3,7 +3,7 @@ import { tokens } from "@equinor/eds-tokens";
 import styled from "styled-components";
 
 export const LockStatusBanner = styled(Banner)`
-  margin-bottom: 1em;
+  margin-bottom: ${tokens.spacings.comfortable.medium};
   border: solid 1px ${tokens.colors.ui.background__medium.hex};
   border-radius: ${tokens.shape.corners.borderRadius};
   box-shadow: none;
--- a/frontend/src/components/form/form.style.ts
+++ b/frontend/src/components/form/form.style.ts
@@ -1,16 +1,17 @@
+import { tokens } from "@equinor/eds-tokens";
 import styled from "styled-components";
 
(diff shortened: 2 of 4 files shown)


joargr:
Another place where tokens can be used:

- `frontend/src/routes/index.style.ts`:
   - `AppContainer`:
      - `.content`:
         - `padding`: `24px` equals size `large`

PR #191, opened 2026-01-12
ENH: Make button general

### LINKED ISSUE #182: Base `SubmitButton` on `GeneralButton`
The  `GeneralButton` includes code for preventing a button to change width when it is pending. This should be applied to the `SubmitButton` as well and the best is to base the `SubmitButton` on the `GeneralButton` to inherit and align their functionality, and to avoid having to update code in two places when new rules are applied.

nourinmohd:
Resolves #182

Based `SubmitButton` and `CancelButton` on `GeneralButton`

## Checklist

### DIFF
--- a/frontend/src/components/form/button.tsx
+++ b/frontend/src/components/form/button.tsx
@@ -14,9 +14,9 @@ type GeneralButtonProps = {
 } & Pick<ButtonProps, "variant" | "type">;
 
 export function GeneralButton({
-  label,
+  type,
   variant,
-  type = "button",
+  label,
   disabled,
   isPending,
   tooltipText,


frontend/src/components/form/button.tsx
```diff
@@ -5,24 +5,27 @@ import {
...
+  isPending?: boolean;
+  disabled?: boolean;
+  tooltipText?: string;
+  onClick?: (e: React.MouseEvent<HTMLButtonElement>) => void;
+} & Pick<ButtonProps, "variant" | "type">;
+
 export function GeneralButton({
   label,
   variant,
+  type = "button",
```
joargr: The `type` property is used unchanged as a property on the EDS `Button` component. In this, the default value of `type` is `button`, so I think it's better to rely on this `Button` default value. In other words, the default value on `GeneralButton` should be removed.

PR #192, opened 2026-01-12
ENH: Align page title to sumo

### LINKED ISSUE #58: Align page title with Sumo
We'd like a consistent user experience through the FMU pages. Currently, the only example of web FMU application is Sumo. We should align the page title with the style used there, with the idea that if you were to click a link from FMU Settings to the Sumo UI, it appears fairly seamless.

Replace:

https://github.com/equinor/fmu-settings-gui/blob/f2fe5f293da0a5ce64776da8965335338d517580/frontend/src/components/Header.tsx#L35

With a `PageTitle` (or similar) component as in Sumo UI:

```jsx
<PageTitle>FMU Settings</PageTitle>
```

```js
export const PageTitle = styled.span`
	white-space: nowrap;
	margin-right: 16px;

	@media (max-width: 800px) {
		display: none;
	}
`;
```

Ref: https://github.com/equinor/sumo-ui-fmu/tree/main/src/components/shared/Header

nourinmohd:
Resolves #58 

Aligned page title with sumo. 

## Checklist


frontend/src/components/Header.tsx
```diff
@@ -170,7 +169,9 @@ export function Header() {
           >
             <FmuLogo src={fmuLogo} />
           </Button>
-          <Typography variant="h1_bold">FMU Settings</Typography>
+          <Typography style={{ marginRight: "16px", whiteSpace: "nowrap" }}>
+            FMU Settings
+          </Typography>
```
tnatt: Initially thought is that the text looks really small, but it is equal to the sumo ui.  And it seems the standard size for the EDS TopBar.

The `marginRight` is not needed here since we don't have any element to the right of the title, and I am not sure we will ever, so maybe just remove it. Also we probably don't need the `whiteSpace: "nowrap"` since our title is so short🙂 

Actually doing this is giving the same style as the code you have. Seems the `TopBar.Header` applies styling for the text.
```suggestion
          FMU Settings
```

But no harm in including it of course. If we are going to style the title explicitly, it would be nice to store styles for the title in a PageTitle component in common/styles. to gather styles in a common place and avoid specifying the style directly here.

nourinmohd:
This is how it looks.  I did not add the Ert link since it only points to the Ert documentation. But let me know if it should be added. 


Screenshot 2026-02-24 at 09 04 58
Screenshot 2026-02-24 at 09 05 05

frontend/src/components/Header.tsx
```diff
@@ -162,15 +163,19 @@ export function Header() {
...
+            <AppMenu />
+            <Button
+              variant="ghost"
+              as={Link}
+              to="/"
+              style={{ backgroundColor: "inherit" }}
+            >
+              <FmuLogo src={fmuLogo} />
+            </Button>
+            <Typography>FMU Settings</Typography>
```
GibranAlfa: this should use h1_bold

GibranAlfa (commented):
looking good, i think we dont need to show ert to align with sumo and webviz, thus ert icon should be removed, as well as other png logos that are unused

PR #193, opened 2026-01-12
BUG: Remove default maxWidth for EditDialog

### LINKED ISSUE #132: Don't use default maxWidth on edit dialog

nourinmohd:
Resolves #132 and #253. 

Added a `GenericDialog `component and extended the padding bottom to `EditDialog`. 

## Checklist

### DIFF
--- a/frontend/src/components/project/masterdata/Edit.tsx
+++ b/frontend/src/components/project/masterdata/Edit.tsx
@@ -142,7 +142,7 @@ function ConfirmItemsOperationDialog({
   }
 
   return (
-    <EditDialog open={isOpen} $minWidth="32em" $extraPaddingBottom={false}>
+    <EditDialog open={isOpen} $minWidth="32em">
       <Dialog.Header>
         <Dialog.Title>
           {selectedItems.operation === "addition" ? "Add" : "Remove"} items
--- a/frontend/src/styles/common.ts
+++ b/frontend/src/styles/common.ts
@@ -103,29 +103,30 @@ export const InfoChip = styled(Chip)`
   }
 `;
 
-export const EditDialog = styled(Dialog).attrs<{
+export const GenericDialog = styled(Dialog).attrs<{
   $minWidth?: string;
   $maxWidth?: string;
-  $extraPaddingBottom?: boolean;
-}>((props) => ({
+}>(({ $minWidth = "10em", $maxWidth = undefined }) => ({
   style: {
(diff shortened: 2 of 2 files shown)


frontend/src/components/Header.tsx
```diff
@@ -110,7 +110,7 @@ function FeedbackDialog() {
         isDismissable={true}
         open={isOpen}
         onClose={closeDialog}
-        $maxWidth="32em"
+        style={{ maxWidth: "32em" }}
```
nourinmohd: I am not sure if this is correct, there are other places that also needs to be changed that uses the EditDialog component. If this is the right way, then I'll make another commit with the respective changes.

joargr:
The `EditDialog` component now has a new attribute `extraPaddingBottom`, with a default value of `true`. The definition of this should be updated as well, as described in the issue.

PR #194, opened 2026-01-13
ENH: Add option to set RMS stratigraphy

### LINKED ISSUE #176: Add functionality for setting RMS stratigraphy in the project
We need to implement fetching stratigraphic data from RMS and store the stratigraphy (horizons/zones) to the project.
The user should be able to filter out from the available stratigraphic data what they want to store.

tnatt:
Resolves #176 

PR to add functionality for displaying and editing the project stratigraphy from available stratigraphy in RMS. This includes adding a new component for visualizing a stratigraphic framework.

For now, while waiting to get stratigraphic column information from RMS, the PR includes functionality for sorting overlapping zones into separate stratigraphic columns. The column information will be possible to extract from RMS projects that are opened in RMS version >15, and PR #220 will enable us to extract this for all models.  

Currently project stratigraphy that is not present in available (orphan) needs to be removed before new data can be stored. In a future PR we should add the possibility to remap an orphan project item to an available item, since a common use case is that a zone or horizon have been renamed in RMS. Then existing stratigraphy mappings to official data can be updated as well with the new name, instead of having to be redone.  

image

image

## Checklist


frontend/src/components/project/rms/Stratigraphy.tsx
```diff
@@ -0,0 +1,435 @@
...
+          </List.Item>
+        </List>
+      </div>
+    </Banner>
+  );
+}
+
+function StratigraphyEditor({
+  projectHorizons,
+  projectZones,
```
joargr: Do these two values need to be transferred as parameters? Is it not possible to get the values from `form`, which is already being used for other purposes? I think it would be good to reduce the number of parameters from four to two.

frontend/src/components/project/rms/Stratigraphy.tsx
```diff
@@ -0,0 +1,435 @@
...
+  const orphanHorizonNames = namesNotInReference(
+    projectHorizons,
+    availableHorizons,
+  );
+  const hasOrphans =
+    orphanHorizonNames.length > 0 || orphanZoneNames.length > 0;
+
+  form.setErrorMap({
+    onChange: hasOrphans ? ["Orphan zones or horizons present"] : undefined,
+  });
```
joargr: Sometimes the following error message is shown in the display:

```
Cannot update a component (`LocalSubscribe`) while rendering a different component (`StratigraphyEditor`). To locate the bad setState() call inside `StratigraphyEditor`, follow the stack trace as described in https://react.dev/link/setstate-in-render
```

The stack trace points to this form update as the culprit.

The error has been observed when the FMU project has become read-only, but perhaps it is linked to whenever there are orphans?

frontend/src/components/project/rms/Stratigraphy.tsx
```diff
@@ -0,0 +1,435 @@
...
+              variant="outlined"
+              label="Add all available"
+              disabled={
+                projectHorizons.length === availableHorizons.length &&
+                projectZones.length === availableZones.length
+              }
+              onClick={() => {
+                setConfirmAction("add");
+              }}
+            />
```
joargr: The "Add all available" button is for adding items from the "available" side to the "project" side. As such, it would be natural that the button is placed on the "available" side. The "Remove all" button would remain on the "project" side, as the button is for removing items from the "project" side. When placed on the "available" side, the label should probably be simplified to "Add all", as the location of the button now makes clear what items will be added.

If moving the "Add all available" button to the "available" side, it should probably be placed directly under the items box, and above the tip. While both button and tip relate to the items box, the button is the most essential of those, and thus should be placed closest to the items box.

frontend/src/components/project/rms/Stratigraphy.tsx
```diff
@@ -0,0 +1,435 @@
...
+  return (
+    <EditDialog open={!!confirmAction} $minWidth="max-content">
+      <Dialog.Header>Confirm action</Dialog.Header>
+
+      <Dialog.Content>
+        <PageText $marginBottom="0">
+          {confirmAction === "add"
+            ? "This will add all available stratigraphy to the project."
+            : "This will remove all stratigraphy from the project."}
+          <br />
```
joargr: I don't think there should be an explicit line break here.

frontend/src/components/project/rms/StratigraphicFramework.tsx
joargr: It seems that there is a problem regarding multiple columns showing zones. Using Drogon data, there are two zone columns on the available side, with "TopVolantis_BaseVolantis" shown in the second of these columns and the rest of the zones shown in the first column. When "TopVolantis_BaseVolantis" is added to the project stratigraphy, it is shown as being placed in one zone column, and it's only when another zone, like "Valysar" is added, that the two zones are shown placed in two zone columns. Furthermore, if the second zone is the "Above" zone (instead of "Valysar") then these two zones ("Above" and "TopVolantis_BaseVolantis") is shown placed in one zone column, even though they are shown placed in two zone columns on the available side.

Image

Is this something which should and can be fixed? Ideally, the project side would have two zone columns even when there are zones in only one of the columns.

Another, smaller thing: The padding to the left of the zones is larger than the padding on the right.

frontend/src/services/stratigraphy.ts
```diff
@@ -0,0 +1,104 @@
...
+  const handleAddAll = () => {
+    form.setFieldValue("horizons", availableHorizons);
+    form.setFieldValue("zones", availableZones);
+  };
+
+  return {
+    handleRemoveItems,
+    handleAddItems,
+    handleAddAll,
+    handleRemoveAll,
```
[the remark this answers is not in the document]
tnatt: thanks, did not know that :)

frontend/src/components/project/rms/Stratigraphy.tsx
```diff
@@ -0,0 +1,435 @@
...
+  formComponents: { StratigraphyEditor, CancelButton, SubmitButton },
+});
+
+function ConfirmActionDialog({
+  setConfirmAction,
+  onConfirm,
+  confirmAction,
+}: {
+  onConfirm: () => void;
+  setConfirmAction: React.Dispatch<React.SetStateAction<ConfirmAction>>;
```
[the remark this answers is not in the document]
tnatt: I should have remembered from a previous PR 😳  will update!

frontend/src/components/project/rms/StratigraphicFramework.style.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    &:nth-child(2) {    
+      display: flex;
+      justify-content: center;
+      grid-column: 2 / -1;
+    }
+  }
+`;
+
+export const StratigraphicFrameworkContent = styled.div.attrs<{
```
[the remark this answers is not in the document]
tnatt: nice addition, thanks

frontend/src/components/project/rms/Stratigraphy.tsx
```diff
@@ -0,0 +1,435 @@
...
+            : !isRmsProjectOpen
+              ? "RMS project is not open"
+              : undefined
+        }
+        onClick={() => {
+          setIsDialogOpen(true);
+        }}
+      />
+
+      {isRmsProjectOpen && (
```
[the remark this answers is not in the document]
tnatt: added the `enable` option on the query and removed the check here 👍

frontend/src/components/project/rms/Stratigraphy.tsx
tnatt: yes that seems to be possible, I'll change it!

frontend/src/components/project/rms/Stratigraphy.tsx
```diff
@@ -0,0 +1,435 @@
...
+  onConfirm: () => void;
+  setConfirmAction: React.Dispatch<React.SetStateAction<ConfirmAction>>;
+  confirmAction: ConfirmAction;
+}) {
+  const resetConfirmAction = () => {
+    setConfirmAction("");
+  };
+
+  return (
+    <EditDialog open={!!confirmAction} $minWidth="max-content">
```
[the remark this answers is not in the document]
tnatt: Nice :) that amount of white space have annoyed me

frontend/src/components/project/rms/Stratigraphy.tsx
```diff
@@ -0,0 +1,470 @@
...
+      });
+    },
+    onError: (error) => {
+      if (error.response?.status === HTTP_STATUS_UNPROCESSABLE_CONTENT) {
+        const message = httpValidationErrorToString(error);
+        console.error(message);
+        toast.error(message);
+      }
+    },
+    meta: { errorPrefix: "Error updating project stratigraphy" },
```
joargr: `preventDefaultErrorHandling` should also be specified, to make it the same as other editing mutations.

```suggestion
    meta: {
      errorPrefix: "Error updating project stratigraphy",
      preventDefaultErrorHandling: [HTTP_STATUS_UNPROCESSABLE_CONTENT],
    },
```

frontend/src/components/project/rms/Stratigraphy.tsx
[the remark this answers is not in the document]
joargr: Perhaps this dialog text should be slightly reformatted. In the masterdata editing a similar dialog first spells out the consequences of the action, then shows the confirmation question in a separate paragraph. For the stratigraphy dialog the action consequence is explained in one sentence, but other confirmation dialog might have more information (like the masterdata editing dialog does). Consistently having the confirmation question as a separate paragraph would be good, so I suggest this being done here. The two paragraphs would be set using `PageText`, with the bottom margin on the last one set to 0.

As for the wording, it should be the same among the confirmation dialogs. The masterdata editing dialog uses the wording "Do you want to continue?" (as opposed to "_wish_ to continue"), and I think this is a better phrasing.

PR #231, opened 2026-02-09
Automatically select associated items

### LINKED ISSUE #177: Automatically select associated items when adding to or removing from project
When an item is being added to or removed from a project, any associated items should also be included. What items these should be could be decided by rules:

- If a discovery is added to the project, the associated field should also be added if it's not already part of the project
- If a field is added to the project, the associated country should also be added it it's not already part of the project
- If a field is removed from the project, any discoveries should also be removed
- Possibly other rules as well

As there would be more items added or removed, the user should get a confirmation dialog informing about the changes, with possibility to confirm or cancel the changes.

joargr:
Resolves #177

For any items selected for addition to or removal from a project, checks are made to find any associated items, as described in the issue. When associated items are found, a confirmation dialog is shown allowing the user to confirm or cancel the move operation.

Previously, selecting an item for addition to or removal from a project led to the move operation being immediately performed. With this PR, the flow now is as follows:

- The user selects an item for moving. The item is placed in a list `selectedItems`
- On the next React render, the list `selectedItems` is checked for contents, by way of a `useEffect`. If so, the function `startItemsOperation` is executed
- `startItemsOperation` checks for items that are affected by the move operation, such as a Field when a Discovery is to be added to the project. The function `handleAffectedItems` is called, which in turn calls `checkForAffectedItems` iteratively
- If affected items are found, these are in turn checked for any affected items. A Discovery to be added to an empty project will result in an affected Field item, which in turn is checked and an affected Country item is found. Thus, the addition of a Discovery will result in both a Field and a Country being added as well
- If the affected check results in additional items to be added or removed, a confirmation dialog will be shown. The affected items are kept in the list `affectedItems`, and the selected item remains in `selectedItems`
- When the confirmation dialog shows, it will display a message showing the selected and affected items that will be added to or removed from the project. The user can confirm the move operation, or cancel it
- The move operation is handled by the function `handleItemsOperation`. This is called directly when there are no affected items, or when there is confirmation dialog and the user confirms the move operation
- Finally, `finishItemsOperation` is called to reset the item lists

The main change in this PR is that the selection and move operation are split into separate steps. This allows for checking for affected items, ie. to act on additional items, and for offering a confirmation dialog. It also allows for other ways to select items for moving, in addition to the user selecting an item by clicking on it, such as other functions adding items. Another possibility is to show the confirmation dialog if the user selects multiple items, such as using a "select all discoveries" button. These are features that can be implemented at a later stage.

The code is now split into several files. `Edit.tsx` contains the components (the user interface), whereas the functions and the data processing is stored in `functions.ts`. The types are stored in `types.ts`, and utility functions now are stored in `utils.ts`. Some utilities that can be relevant to other parts of the application are stored in the general `utils` folder.

## Checklist


joargr:
> 1. Brazilian fields do not have coordinate system in SMDA for now, so that will set `smdaMasterdata.isSuccess` to `false` and blocking any subsequent masterdata editing operation. Right now the coordinate system field is set as required in the project config, so by that logic it's okay to have `smdaMasterdata.isSuccess` to check every query to be successful. Not sure by the time the Brazilian field onboarded they will have a coordinate system in SMDA, if not then we should set `smdaMasterdata.isSuccess` to not check on every query results, i.e. let the query for CS fails and still allow operation.

It seems that few if any fields outside of Norway get successful responses today when querying the SMDA API. Trying to add a Brazilian field in the masterdata editing, or a field from another country, only results in a 404 Not found error from the FMU Settings API. Thus, preparing for the onboarding of Brazilian fields will require some work investigating this first.

> 2. #243

This issue is for a bug regarding expiring SSO token used by the SMDA API. Just recently there are changes afoot in how FMU Settings API can communicate with the SMDA API, so for the GUI this is a bug which should soon become irrelevant. I've updated the issue with this information.

> 3. We discussed about validation button in planning today. I'm wondering if we should also validate:
> 
> * Only allow one field in the masterdata on save? Because now we allow multiple field to be in masterdata, e.g. I can add Troll and JS. We allow list of field in the masterdata [model](https://github.com/equinor/fmu-datamodels/blob/8f006c80674caa0a3fe9c9a2f8f6d5f1d9b8f672/src/fmu/datamodels/common/masterdata.py#L35-L37) but does this make sense? Maybe this is for oseberg for.ex. that has multiple field?

I think there are use cases where more than one field is relevant for a project, so OSEBERG and OSEBERG SØR sounds like something that would be relevant.

> * We only allow one stratigraphic column in the [model](https://github.com/equinor/fmu-datamodels/blob/8f006c80674caa0a3fe9c9a2f8f6d5f1d9b8f672/src/fmu/datamodels/common/masterdata.py#L39). Do we also want this? Some field have many strat column, like Troll has Troll and Troll West. In that case they should do fmu settings separately?

If there needs to be multiple values for stratigraphic column stored in a project, that should first be reflected in the model. Perhaps this could be a topic for discussion, and if needed, the GUI could be adjusted to allow for multiple values.

PR #252, opened 2026-02-18
ENH: Add lock expire dialog

### LINKED ISSUE #186: Add `Lock about to expire` dialog
Currently the project lock expires without any warnings. We should introduce a dialog pop up some minutes before the lock is about to expire to give the users an option to extend the lock.

nourinmohd:
Resolves #186 

Add lock expire dialog. 

## Checklist

### DIFF
--- a/frontend/src/components/LockExpireNotification.tsx
+++ b/frontend/src/components/LockExpireNotification.tsx
@@ -61,19 +61,22 @@ export function LockExpireNotification() {
       return;
     }
 
-    const updateTimeUntilExpire = () => {
-      const lockExpireAt = new Date(lockInfo.expires_at * 1000).getTime();
-      const timeLeft = Math.max(
-        0,
-        Math.ceil((lockExpireAt - Date.now()) / 1000) * 1000,
-      );
+    const initialTimeLeft = Math.max(
+      0,
+      Math.ceil(lockInfo.expires_at - Date.now() / 1000),
+    );
 
-      setTimeUntilExpire(timeLeft);
-    };
+    setTimeUntilExpire(initialTimeLeft);
 
-    updateTimeUntilExpire();
+    const interval = setInterval(() => {
+      setTimeUntilExpire((currentTimeLeft) => {
+        if (!Number.isFinite(currentTimeLeft) || currentTimeLeft <= 0) {
+          return currentTimeLeft;
+        }
 
-    const interval = setInterval(updateTimeUntilExpire, 1000); // setInterval : countdown
+        return currentTimeLeft - 1;
+      });
+    }, 1000);
 
     return () => {
       clearInterval(interval);
@@ -96,8 +99,6 @@ export function LockExpireNotification() {
       timeUntilExpire !== Number.POSITIVE_INFINITY
     ) {
       setTimeUntilExpire(Number.POSITIVE_INFINITY);
-    } else {
-      return;
     }
   }, [isDialogOpen, isLockAcquired, timeUntilExpire, queryClient]);
 
@@ -125,7 +126,7 @@ export function LockExpireNotification() {
           <>
             <PageText>
               Your lock will expire and the project will become read-only in{" "}
-              <b>{Math.ceil(timeUntilExpire / 1000)}</b> seconds.
+              <b>{timeUntilExpire}</b> seconds.
             </PageText>
 
             <PageText $marginBottom="0">
--- a/frontend/src/config.ts
+++ b/frontend/src/config.ts
@@ -13,8 +13,8 @@ export const ssoScopes = [
   "691a29c5-8199-4e87-80a2-16bd71e831cd/user_impersonation", // SMDA
 ];
 
-export const projectLockStatusRefetchInterval = 60_000; // 1 minute
+export const projectLockStatusRefetchInterval = 60; // 1 minute
 
-export const projectLockExpireNotificationThreshold = 120_000; // 2 minutes
+export const projectLockExpireNotificationThreshold = 120; // 2 minutes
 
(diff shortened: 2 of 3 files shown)


frontend/src/components/LockExpireDialog.tsx
```diff
@@ -0,0 +1,132 @@
...
+  projectGetLockStatusQueryKey,
+  projectPostLockRefreshMutation,
+  projectPostLockReleaseMutation,
+} from "#client/@tanstack/react-query.gen";
+import { projectLockTimeoutWarningThreshold } from "#config";
+import { EditDialog, PageText } from "#styles/common";
+
+export function LockExpireDialog({ lockStatus }: { lockStatus: LockStatus }) {
+  const [isDialogOpen, setIsDialogOpen] = useState(false);
+  const [timeUntilExpire, setTimeUntilExpire] = useState<number>(200000);
```
GibranAlfa: as this is only a placeholder, we can use
```suggestion
  const [timeUntilExpire, setTimeUntilExpire] = useState<number>(Number.POSITIVE_INFINITY);
```

frontend/src/components/LockExpireDialog.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  const lockInfo = lockStatus.lock_info;
+  const isExpired = !lockStatus.is_lock_acquired;
```
GibranAlfa:
```suggestion
  const isLockAcquired = lockStatus.is_lock_acquired;
  const isExpired = !isLockAcquired;
```

GibranAlfa (commented):
Currently `LockExpireDialog` can surface for sessions that doesn't hold the lock, because we only check if there's a lock info. We should show the dialog only if the session hold the lock.

frontend/src/components/LockExpireDialog.tsx
```diff
@@ -0,0 +1,132 @@
...
+            onClick={() => {
+              setIsDialogOpen(false);
+            }}
+          >
+            Close
+          </Button>
+        ) : (
+          <>
+            <Button onClick={onLockRefresh}>Extend lock</Button>
+            <Button onClick={onLockRelease}>Release lock</Button>
```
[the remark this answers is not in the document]
GibranAlfa: depends on the intention. i think we want the user to extend the lock with this dialog, so we shift the focus on extend lock button. if they are equally important then we can have the buttons look the same

frontend/src/components/Header.tsx
```diff
@@ -64,30 +65,36 @@ function ProjectInfo() {
...
-          <ProjectInfoItem
-            label="Revision"
-            value={project.data.config.model?.revision}
-          />
-        </>
-      ) : (
-        "No project selected"
+    <>
+      {project.lockStatus && (
+        <LockExpireDialog lockStatus={project.lockStatus} />
```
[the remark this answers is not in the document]
joargr: As this dialog can appear on any page, whether a project page, a user page or on the home page, the component should be placed somewhere central. Putting it in `routes/__root.tsx` thus makes sense. And this is where the `ToastContainer` component is placed, which has a similar purpose.

frontend/src/components/Header.tsx
```diff
@@ -23,7 +24,6 @@ import {
   ProjectInfoItemContainer,
   TopBarContainer,
 } from "./Header.style";
-import { AppMenu } from "./home/AppMenu";
```
GibranAlfa: why AppMenu is removed?

frontend/src/components/LockExpireNotification.tsx
```diff
@@ -0,0 +1,141 @@
...
+          <Button
+            onClick={() => {
+              setIsDialogOpen(false);
+            }}
+          >
+            Close
+          </Button>
+        ) : (
+          <>
+            <Button onClick={onLockRefresh}>Extend lock</Button>
```
joargr: The `GeneralButton` component builds on the EDS `Button` component and adds functionality. I think as a general pattern the application should use `GeneralButton` for buttons, as it allows for easily adding functionality such as tooltips and disabled state at a later stage, should that be necessary.

frontend/src/components/LockExpireNotification.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+    const updateTimeUntilExpire = () => {
+      const lockExpireAt = new Date(lockInfo.expires_at * 1000).getTime();
+      const timeLeft = Math.max(
+        0,
+        Math.ceil((lockExpireAt - Date.now()) / 1000) * 1000,
+      );
+
+      setTimeUntilExpire(timeLeft);
+    };
```
joargr: This code is still prone to skipping values in the countdown. Testing with a 60 s countdown, it typically skips 1 value (ie. going "31 30 28 27"). This looks buggy to the user, so I think it is better to have a countdown that doesn't skip values, even though that might result in the lock going away before the countdown actually reaches 0 (it would happen perhaps when the countdown is at 1 or 2). Having a countdown that counts precisely is better than a countdown that reaches 0 at the exact moment the lock goes away, as I think most users will take action (keep the lock or release the lock) only a few seconds after the dialogs becomes visible. So few if any users will actually observe the lock going away and the countdown being at 1 or 2, and if the countdown completes then that is probably a situation where the user isn't using the application or looking at the page at that moment.

Suggest calculating the initial `timeUntilExpire` value first, based on `lockInfo.expires_at`, and storing this, then starting the interval where the interval function simply decreases the value by 1 s on each interval. Whenever `lockInfo` is updated `timeUntilExpire` will be recalculated, and a new interval countdown will be started. It should also be enough to keep track of `timeUntilExpire` as a value in whole seconds, instead of in milliseconds.

PR #254, opened 2026-02-19
ENH: Add project history page

### LINKED ISSUE #229: Add a revision management page
This page will present the project history by showing the list of revisions from newest to oldest and add the changelog viewer in between them to see what has been changed since the previous revision (cache). If we click view details, we will show the diff of current file and the selected revision (for now we show the whole content because no endpoint to show the diff only). If we click restore, it will overwrite current file to the selected revision.

It will look something like this (mock up)

Image

Image

- I will make another endpoint to only view the diff (https://github.com/equinor/fmu-settings-api/issues/280)
- The changelog view is using a mock data as we don't have the endpoint for that yet.

Question: I see that the content area has `max-width: 55em`, can we expand this or is there any reasoning behind this? Would be nice to use that empty space on the right.

What do you guys think on this plan? @tnatt @joargr @nourinmohd @mferrera 

We can also add a full changelog viewer on the right side to see the full project history because the cache only show 5 latest revisions.

### LINKED ISSUE #250: Width mismatch between EdsSideBar and AppContainer
EDS SideBar component has a fixed width of 256 px (because we always set it to open), but the grid column in https://github.com/equinor/fmu-settings-gui/blob/f042e1e518570df1209c8cdc860fc0f9832d8cf5/frontend/src/routes/index.style.ts#L6 uses min-content, which can cause width mismatch during loading states.

Image

GibranAlfa:
Resolves #229 
Resolves #250 

Added project history page so users can restore resources (config and mappings for now) from cache. See video below for example (video is before the bug fix that can have stale diff and stale current state after restore):

https://github.com/user-attachments/assets/695b5a96-df86-4595-994a-d230670489bf

When we restore a resource, the current state are saved as a new cache and then overwritten with the selected snapshot. If the user wants to go back to previous state (pre-restore), they can restore it from the pre-restore.

Added/removed/updated is what will happen if the user click restore (e.g. field will be added if they restore it with selected snapshot)

## Checklist

### DIFF
--- a/frontend/src/components/project/history/Viewer.tsx
+++ b/frontend/src/components/project/history/Viewer.tsx
@@ -29,7 +29,7 @@ import {
   PageText,
 } from "#styles/common";
 import { ReadableValue } from "./ReadableValue";
-import type { CacheEntry } from "./types";
+import type { CacheEntry, DiffKind } from "./types";
 import {
   formatCacheDateTime,
   formatFieldPath,
@@ -48,7 +48,6 @@ import {
   DiffDialogContent,
   DiffFieldHeader,
   DiffGroup,
-  type DiffKind,
   DiffLegend,
   ResourcePickerContainer,
   SnapshotInfo,
--- a/frontend/src/components/project/history/Viewer.style.ts
+++ b/frontend/src/components/project/history/Viewer.style.ts
@@ -2,8 +2,7 @@ import { tokens } from "@equinor/eds-tokens";
 import styled from "styled-components";
 
 import { GenericBox, GenericInnerBox, InfoBox } from "#styles/common";
-
-export type DiffKind = "added" | "removed" | "updated";
+import type { DiffKind } from "./types";
 
 const diffPalette: Record<DiffKind, { background: string; color: string }> = {
   added: {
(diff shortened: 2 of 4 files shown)


GibranAlfa:
Changes after review:

1. The "Restore" button has been removed from the snapshot list. Users must click "View details" to open the diff dialog before they can restore.
2. Restore is disabled in the diff dialog when there's nothing to restore (no diff)
3. No diff legend if there's no diff
4. Snapshot naming is consistent after a restore, now labelled "Snapshot 1 (pre-restore)"
5. Combine data and time to single line

frontend/src/components/project/history/Viewer.tsx
```diff
@@ -0,0 +1,543 @@
...
+        current version and a snapshot, and to <strong>restore</strong> from it.
+      </PageText>
+      {projectReadOnly && (
+        <PageText $marginBottom="0">
+          Project is read-only. Restore is disabled until the project is opened
+          for editing.
+        </PageText>
+      )}
+
+      <div style={{ maxWidth: "15em" }}>
```
joargr: This container should ideally be a styled component, so as to avoid setting style directly.

PR #269, opened 2026-02-25
ENH: Set max snapshots and recover .fmu files in History page

### LINKED ISSUE #255: Set max cache and restore from memory in project history
depends on:

- [x] https://github.com/equinor/fmu-settings-api/issues/300

- [x] https://github.com/equinor/fmu-settings-api/issues/195

### LINKED ISSUE #333: Improve the term "resource" in the history page to something user friendly
The current sentence, "Choose a resource to view its snapshots, listed from newest to oldest," could be improved for clarity and user-friendliness. The term "resource" may be too generic and unclear for users.

An alternative term considered is "settings-type," but this may also cause confusion, as users might be uncertain about what specific settings type they are selecting.

GibranAlfa:
Resolves #255 
Resolves #333 

- Added Max snapshots dropdown to set max cache (limited to 5, 10, 15, 20). Save button is disabled if the value is the same as the current max snapshots
- Added "Recover project/user files" section to restore from in-memory state
- Use `ScrollableCardStack` for the snapshot lists for many snapshot lists
- Split the old project history viewer into smaller components: `Overview`, `SnapshotHistory`, `ProjectFileRecovery`
- Renamed "Resource" into "settings type"
- Added User sidebar entry for Recovery

## Checklist

### DIFF
--- a/frontend/src/components/Sidebar.tsx
+++ b/frontend/src/components/Sidebar.tsx
@@ -25,6 +25,7 @@ export function Sidebar() {
     ProjectSubItems.push({ label: "Stratigraphy", to: "/stratigraphy" });
     ProjectSubItems.push({ label: "History", to: "/history" });
   }
+  ProjectSubItems.push({ label: "History", to: "/project/history" });
 
   return (
     <EdsSideBar open>
--- a/frontend/src/components/Sidebar.tsx
+++ b/frontend/src/components/Sidebar.tsx
@@ -70,17 +70,17 @@ export function Sidebar() {
           isExpanded={userExpanded}
         >
           <EdsSideBar.AccordionItem
-            label="Recovery"
-            as={Link}
-            to="/user/recovery"
-            active={currentPath === "/user/recovery"}
-          />
-
-          <EdsSideBar.AccordionItem
             label="API keys"
             as={Link}
             to="/user/keys"
             active={currentPath === "/user/keys"}
+          />
+
+          <EdsSideBar.AccordionItem
+            label="Recovery"
+            as={Link}
+            to="/user/recovery"
+            active={currentPath === "/user/recovery"}
           />
         </EdsSideBar.Accordion>
       </EdsSideBar.Content>
--- a/frontend/src/components/Sidebar.tsx
+++ b/frontend/src/components/Sidebar.tsx
@@ -25,7 +25,6 @@ export function Sidebar() {
     ProjectSubItems.push({ label: "Stratigraphy", to: "/stratigraphy" });
     ProjectSubItems.push({ label: "History", to: "/history" });
   }
-  ProjectSubItems.push({ label: "History", to: "/project/history" });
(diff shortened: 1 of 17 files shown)


GibranAlfa:
Changes after review:

- Added description of that Max snapshots does
- Renamed "Resource" into "Settings type"
- Added confirmation dialog before lowering max snapshots when old snapshots will be deleted.
- Added deleted-file recovery flow:
  - project file recovery on project history page
  - user file recovery on new `/user/recovery` page
  - shared dialog that first checks for deleted files, then only shows Recover when something can be recovered
- Immediately trim the snapshots if the new max snapshots is lower than current snapshots (through backend change)
- Split the old project history viewer into smaller components: `Overview`, `SnapshotHistory`, `ProjectFileRecovery`
- Fix settings type picker width to show the full resource name on TGX
- Use (not set) dummy value instead of Select and not call API on that
- Centralized hardcoded React Query key IDs in `frontend/src/utils/query.ts.`

frontend/src/components/project/history/SnapshotHistory.tsx
```diff
@@ -361,23 +381,110 @@ function RestoreSnapshotDialog({
...
+      </Dialog.Header>
+
+      <Dialog.Content>
+        <PageText>
+          Reducing max snapshots to
+          <span className="emphasis">
+            {" "}
+            {maxRevisions !== undefined ? String(maxRevisions) : "this value"}
+          </span>{" "}
+          will delete
```
joargr: I think the phrasing should be "...will delete **the** n oldest snapshots...", ie. add the word "the".

frontend/src/components/Sidebar.tsx
```diff
@@ -69,6 +69,13 @@ export function Sidebar() {
           icon={account_circle}
           isExpanded={userExpanded}
         >
+          <EdsSideBar.AccordionItem
+            label="Recovery"
+            as={Link}
+            to="/user/recovery"
+            active={currentPath === "/user/recovery"}
+          />
```
joargr, line 77: The new submenu item for recovery of deleted files should be placed after the API keys item. Recovery is a feature that I imagine wouldn't be needed that often, so it should have less emphasis in the menu.

frontend/src/components/project/history/SnapshotHistory.tsx
```diff
@@ -41,18 +67,12 @@ import {
...
-  DiffDialogContent,
-  DiffFieldHeader,
-  DiffGroup,
-  DiffLegend,
-  ResourcePickerContainer,
-  SnapshotInfo,
-  ValuePanel,
-} from "./Viewer.style";
+
+type SnapshotDeletionImpact = {
```
joargr: I guess this belongs in `components/project/history/types.ts`.

joargr (approved):
This looks good, and provides nice functionality.

PR #277, opened 2026-02-27
BUG: Select button enable for read only project path

### LINKED ISSUE #270: Cannot set project path if no project is selected
We cannot set project path if no project is selected because the select button is disabled (it's in project read only mode). This select button should be enabled when we don't have a project, so the user can specify path to their project.

Image

nourinmohd:
Resolves #270 

The button is enabled  even when the project is read only and the user put the project path. 

## Checklist

### DIFF
--- a/frontend/src/components/project/overview/ProjectSelector.tsx
+++ b/frontend/src/components/project/overview/ProjectSelector.tsx
@@ -55,11 +55,9 @@ const { useAppForm: useAppFormProjectSelectorForm } = createFormHook({
 type ValueSource = "recentProjectPath" | "projectPath" | "";
 
 function ProjectSelectorForm({
-  projectReadOnly,
   closeDialog,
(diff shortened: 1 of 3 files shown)


tnatt:
> very nice! this already solve the bug i mentioned. however, a cleaner way to do this is to remove `projectReadOnly` altogether from disabling the Select button. This button is used to switch/input a path to a project. There is no reason to disable it if the project is read only. Project read only should only prevent writing to a project, not switching project. I guess the intention for this is to prevent writing project path to `recent_project_directories` in user .fmu, but that doesn't require project lock, so the logic is flawed.
> 
> I'll let @tnatt confirm this

I confirm and agree with @GibranAlfa :) this read-only disabling of the select project is a bug and should be removed

GibranAlfa (approved):

PR #297, opened 2026-03-10
ENH: Enable asset selection from list

### LINKED ISSUE #265: Use `project/sumo_assets` endpoint to show valid sumo assets
When the user wants to fill in the Sumo target asset, valid sumo assets should be presented as option (maybe with Autocomplete) so it can be selected by the user.

tnatt:
Resolves #265 

## Checklist

### DIFF
--- a/frontend/src/components/project/overview/Access.tsx
+++ b/frontend/src/components/project/overview/Access.tsx
@@ -8,22 +8,23 @@ import {
   Typography,
 } from "@equinor/eds-core-react";
 import { createFormHook } from "@tanstack/react-form";
-import { useMutation, useQueryClient } from "@tanstack/react-query";
+import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
 import { useState } from "react";
 import { toast } from "react-toastify";
 
 import type { FmuProject } from "#client";
 import {
   projectGetProjectQueryKey,
+  projectGetSumoAssetsOptions,
   projectPatchAccessMutation,
 } from "#client/@tanstack/react-query.gen";
-import type { Access, Classification } from "#client/types.gen";
+import type { Access, Classification, SumoAsset } from "#client/types.gen";
 import {
   CancelButton,
   GeneralButton,
   SubmitButton,
 } from "#components/form/button";
-import { TextField } from "#components/form/field";
+import { AutocompleteField, TextField } from "#components/form/field";
 import {
   EditDialog,
   InfoBox,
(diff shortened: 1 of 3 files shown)


nourinmohd (commented):
Gave it a quick shot. Works well, but when playing around with it I got a bit confused why the alternate box was disabled when I was choosing from the list and vice versa.  But otherwise did not manage to break it 👍. 
Also, I wonder if the user enters the name manually and it is not onboarded yet can then manually entered asset then automatically be added to the list ?

frontend/src/components/project/overview/Access.tsx
```diff
@@ -89,17 +95,30 @@ function AccessEditorForm({
     },
   });
 
+  const { data: sumoAssets } = useSuspenseQuery(projectGetSumoAssetsOptions());
```
GibranAlfa: `sumo_assets` is only used in this dialog, so using `useSuspenseQuery` here means opening the dialog can end up suspending the whole project page if that request is slow or fails. I tried this by making the `sumo_assets.json` invalid in the API.

It would be safer to use a normal `useQuery` here and keep the loading/error handling inside the dialog instead.

frontend/src/components/project/overview/Access.tsx
```diff
@@ -114,7 +133,7 @@ function AccessEditorForm({
   });
 
   return (
-    <EditDialog open={isDialogOpen}>
+    <EditDialog open={isDialogOpen} $minWidth="25em">
```
joargr: The `$minWidth` value set on the Model dialog is `32em`. I think it would be good if the two dialogs for setting information would have the same width. Suggest changing the `$minWidth` value of the Model dialog to `25em`, the same as the Access control dialog.

frontend/src/components/project/overview/Access.tsx
```diff
@@ -127,12 +146,38 @@ function AccessEditorForm({
...
+          <form.Subscribe selector={(state) => [state.values]}>
+            {([formValues]) => (
+              <>
+                <form.AppField name="assetName">
+                  {(field) => (
+                    <field.AutocompleteField
+                      label="Select sumo target asset"
+                      options={sumoAssets
+                        .map((asset) => asset.name)
+                        .sort((a, b) => a.localeCompare(b))}
```
joargr: The `stringCompare` util function should be used for sorting, as it uses the Norwegian locale. This is the sorting used in the application, and it results in a field like "Aasta Hansteen" is sorted under "Å".

PR #306, opened 2026-03-16
ENH : Add Changelog view to homepage 

### LINKED ISSUE #281: Add changelog view to the homepage
We want the users to see the changes that has been made since the last time the project was opened. The first task is to add the changelog view to the homepage.

nourinmohd:
Resolves #281 

I've added the `Changelog` view to the homepage.  Check out the latest commit. 

Screenshot 2026-03-16 at 06 48 14

Screenshot 2026-03-16 at 06 48 54

## Checklist


GibranAlfa (commented):
it's looking good visually, i like it. some comments to consider below.

additional comment: we should also invalidate onSuccess in ProjectRecoveryNotification. Right now there's no new changelog entry when we restore, so this is effectively not an issue now, but I think it's better to log it to changelog and we can make that change on the backend.

GibranAlfa (commented):
still need to invalidate queries in onSuccess of restoreFromLatestCache mutation in ProjectRecoveryNotification

frontend/src/components/home/ChangeLog.tsx
```diff
@@ -0,0 +1,129 @@
...
+}
+
+export function ChangeLog() {
+  const { data, isPending, isError, error } = useQuery({
+    ...projectGetChangelogOptions(),
+    meta: { preventDefaultErrorHandling: [404] },
+    retry: (failureCount, queryError) =>
+      !(isAxiosError(queryError) && queryError.response?.status === 404) &&
+      failureCount < 3,
+  });
```
joargr: Suggest using a similar structure as the masterdata and RMS pages, with a suspense query and an error boundary. Using such a structure will automatically take care of the pending and error states, and will reduce the complexity of the code here.

The code would then be something like the following:

```typescript
function Content() {
  const { data } = useSuspenseQuery({
    ...projectGetChangelogOptions(),
    meta: { preventDefaultErrorHandling: [404] },
  });

  if (data.length === 0) {
    <PageText>No changelog entries yet.</PageText>
  }

  // Rest of the code
}

export function Changelog() {
  return (
    <>
      <PageHeader $variant="h3">Changelog</PageHeader>

      <QueryErrorBoundary
        messages={{ 404: "No changelog found for this project." }}
      >
        <Suspense fallback={<Loading />}>
          <Content />
        </Suspense>
      </QueryErrorBoundary>
    </>
  );
}
```

The page header is defined once, and will be present for all states. Note that this change depends on the new [error boundary component](https://github.com/equinor/fmu-settings-gui/pull/319) being added to the application.

src/fmu_settings_gui/static/.gitkeep
joargr, line 1: I don't think this file is meant to be deleted for this issue. This empty file needs to be here so that the `static` directory is created and exists, as the directory where the built React application is placed and served to the browser.

GibranAlfa: this `.gitkeep` is always deleted when you run `pnpm build` (`"build": "tsc -b && vite build --outDir ../src/fmu_settings_gui/static --emptyOutDir"`). that `--emptyOutDir` deletes the `.gitkeep`, so I always have to exclude that from commits. Maybe we should modify the command somehow to not delete `.gitkeep`?

nourinmohd: I didn't even notice it!!

GibranAlfa, line 1: just remove this file from your commit

joargr: When executing the building of the frontend application through the Python setup, the `.gitkeep` file is restored after the Python setup runs the `pnpm` build command (as per [setup.py](https://github.com/equinor/fmu-settings-gui/blob/0acd8b83130a17933fccdc02265fcbcf0c8b63b4/setup.py#L28-L34)). So it seems that the proper way to build the frontend application is through the Python setup, and I guess that's how the CI does things.

I don't think that it is necessary to build the frontend application when developing, but if so the building should happen through the Python setup, as this is what CI does.

frontend/src/components/home/Changelog.tsx
```diff
@@ -0,0 +1,134 @@
...
+        <PageHeader $variant="h3">Changelog</PageHeader>
+        <PageText>No changelog entries yet.</PageText>
+      </>
+    );
+  }
+
+  const latestChanges = [...data]
+    .sort(
+      (a, b) => timestampToNumber(b.timestamp) - timestampToNumber(a.timestamp),
+    )
```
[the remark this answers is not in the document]
joargr: For some reason, the type in the API code results in a TypeScript type of [`timestamp?: string`](https://github.com/equinor/fmu-settings-gui/blob/main/frontend/src/client/types.gen.ts#L90-L98), so the value is treated as being possibly `undefined`. The reference to `timestamp` should therefore have a fallback of the empty string.

@nourinmohd: Suggest that the sorting function be simplified to use `localeCompare` directly, which would mean that the function `timestampToNumber` is no longer needed.

frontend/src/components/home/utils.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  return `${concise.slice(0, 69)}...`;
+}
+
+export function formatEntryDescription(entry: {
+  key: string;
+  change_type: ChangeType;
+  change: string;
+}): string {
```
joargr: The type of `entry` given to this function is `ChangeInfo`, so the type in the function declaration should specify that.

```suggestion
export function formatEntryDescription(entry: ChangeInfo): string {
```

frontend/src/components/home/utils.ts
```diff
@@ -0,0 +1,98 @@
...
+
+function getFieldLabel(path: string): string | undefined {
+  if (path in PATH_LABELS) {
+    return PATH_LABELS[path];
+  }
+
+  // Prefix match for paths with sub-keys or array indices (longest match wins)
+  const sortedKeys = Object.keys(PATH_LABELS).sort(
+    (a, b) => b.length - a.length,
+  );
```
joargr: This creates a sorted version of `PATH_LABELS`, and runs possibly every time a changelog entry is displayed. It should be enough to create the sorted version once:

```ts
const PATH_LABELS: Record<string, string> = {
  // data
};

const sortedPathLabels = Object.keys(PATH_LABELS).sort(
  (a, b) => b.length - a.length,
);
```

Then functions would use this sorted version instead of creating their own version on every execution.

frontend/src/components/home/utils.ts
```diff
@@ -0,0 +1,98 @@
...
+  return undefined;
+}
+
+function formatBriefDescription(change: string) {
+  const compact = change.replace(/\s+/g, " ").trim();
+  const withoutDiffPayload = compact
+    .replace(/Old value:\s*[\s\S]*$/i, "")
+    .replace(/New value:\s*[\s\S]*$/i, "")
+    .replace(/\s*->\s*/g, " ")
+    .trim();
```
joargr: When presenting a change that is an update of a stratigraphy mapping, the changelog shows "Updated field 'stratigraphy'. in Mappings", with the period in the middle of the sentence. This is unfortunate. A better way to exclude the diff value and just show the field that has changed is to do something like the following:

```ts
const withoutDiffPayload = compact.replace(/\. Old value:.*/, "");
```

This leaves only the text before the diff value, and also removes the period.

frontend/src/components/home/utils.ts
```diff
@@ -0,0 +1,98 @@
+import type { ChangeType } from "#client/types.gen";
+
+export const FILE_LABELS: Record<string, string> = {
+  "config.json": "Project configuration",
+  "mappings.json": "Mappings",
+};
+
+const PATH_LABELS: Record<string, string> = {
```
joargr: This list of labels only supports keys in the `config.json` file. There should also be support of keys in the `mappings.json` file. Perhaps a second level can be introduced, something like the following:

```ts
const PATH_LABELS: Record<string, Record<string, string>> = {
  "config.json": {
    access: "access control",
    "access.asset": "asset",
  },
  "mappings.json": {
    stratigraphy: "stratigraphy",
  },
};
```

Then there could be separate labels (keys) for each file. With the definition of `stratigraphy` for `mappings.json`, the entry for a change to the stratigraphy mapping would be "Updated stratigraphy in Mappings" instead of the more generic "Updated field 'stratigraphy' in Mappings".

joargr:
There are a few more places where invalidation of the changelog query should be done, so as to force an updated changelog after changes (mutations) have been done in the application:

- `components/project/history/SnapshotHistory.tsx`: `maxRevisionsMutation` in `SnapshotHistory`
- `components/project/stratigraphy/Overview.tsx`: `mappingsMutation` in `Zones`

(These are features that have been implemented after this PR was created.)

src/fmu_settings_gui/static/.gitkeep
joargr, line 1: This file is now on track to be deleted when the changes in this PR are merged to main, and with this file gone the `static` directory won't be created on a fresh install (pull) of the repo. The file needs to stay, so I think it should be enough to just create the file (it is empty), and commit the change to the branch. Then the original deletion of the file in an earlier commit will be undone for when the PR is merged to main.

frontend/src/components/home/Changelog.tsx
```diff
@@ -0,0 +1,102 @@
...
+        {latestChanges.length === 1
+          ? "Showing the most recent change to this project's settings."
+          : `Showing the ${latestChanges.length} most recent changes to this project's settings.`}
+      </PageText>
+
+      <ChangeList>
+        {latestChanges.map((entry) => {
+          return (
+            <ChangeItem
+              key={`${entry.timestamp ?? "no-time"}-${entry.user}-${entry.change_type}-${entry.file}-${entry.path}-${entry.key}`}
```
joargr: Here I think it would be more natural to use a construct like `[entry.timestamp ?? "no-time", entry.user, entry.change_type].join("-")`, since it more clearly shows that it creates a string of values joined with a common character. However, `entry.timestamp` will always have a value, and I think it is highly unlikely that two separate changelog entries (for two different changes) will have the exact same timestamp, down to the microsecond. Therefore it should be enough to use just the `timestamp` value for creating a unique key value for the React list of change items.

frontend/src/components/home/ChangeLog.tsx
```diff
@@ -0,0 +1,139 @@
...
+}
+
+function getTypeLabel(changeType: string) {
+  return toTitleCase(changeType);
+}
+
+export function ChangeLog() {
+  const { data, isPending, isError, error } = useQuery(
+    projectGetChangelogOptions(),
+  );
```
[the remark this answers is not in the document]
GibranAlfa: this is not yet addressed, you still retry on 404, pls check again if all comments are addressed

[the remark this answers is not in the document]
nourinmohd: Implemented.

frontend/src/components/project/overview/ProjectSelector.tsx
```diff
@@ -98,6 +99,9 @@ function ProjectSelectorForm({
       void queryClient.invalidateQueries({
         queryKey: projectGetProjectQueryKey(),
       });
+      void queryClient.invalidateQueries({
```
GibranAlfa: This should reset/remove the changelog query, not only invalidate it. Since the changelog query key is not project specific, switching to a different project will leave stale entries from the previous project visible on the homepage. You can try this yourself. So, we should use `queryClient.resetQueries` instead of `queryClient.invalidateQueries` in all the code you added.

frontend/src/components/home/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  return undefined;
+}
+
+function formatBriefDescription(change: string) {
+  const compact = change.replace(/\s+/g, " ").trim();
+  const withoutDiffPayload = compact
+    .replace(/\. Old value:.*/i, "")
+    .replace(/\s*->\s*/g, " ")
```
joargr: The previous `replace` will remove `. Old value:` and everything after that, including the ` -> ` part, so this `replace` is not necessary. The `i` flag is not needed, as the text from the API response is exactly "Old value".  Also, it's not necessary to `trim` the string as there won't be any whitespace at the end and the response from the API is known to not have whitespace at the beginning.

frontend/src/components/project/overview/ProjectSelector.tsx
nourinmohd: Screenshot 2026-06-01 at 09 44 54
Screenshot 2026-06-01 at 09 44 44

Gave this a shot with the current implementation of `queryClient.invalidateQueries`, seems the changelog information changes when I change the project.

GibranAlfa: can you try changing from a project that has .fmu and has changelog to a new project just initialized and has no changelog?

nourinmohd: Screenshot 2026-06-01 at 10 18 05

Just created this new_project now, so no changelog . Switched between another project that has changelog and got the changelog of that project ( ..../test_data/project_1)

Screenshot 2026-06-01 at 09 44 44

when switched again to the new_project, no changelog was shown since it did not have.

GibranAlfa: strange, i have the new project showing the stale changelog from previous project, video attached

https://github.com/user-attachments/assets/d033b4e2-ceae-43f8-9f33-14ba9999d3e0

joargr: I guess there are some nuances in the various query managing functions (invalidate/refetch/reset), but with invalidate query perhaps you can try specifying [`refetchType: all`](https://tanstack.com/query/latest/docs/reference/QueryClient#queryclientinvalidatequeries) and see if the changelog is updated as expected to match the newly selected project.

GibranAlfa: Can you reproduce the same thing Joar? Or it's just fine in your computer?

joargr: I can reproduce what @GibranAlfa sees. When I switch from a project that has a changelog to a new project that doesn't have a changelog, the changelog from the previous project is shown. As the changelog query is invalidated when switching the project, a new fetch of the changelog is done when showing the new project. However, this fetching fails with a 404 error, and then the contents of the (currently invalidated) query is shown, ie. the changelog of the previous project. Doing `resetQueries` instead of `invalidateQueries` when switching project leads to the contents of the previous changelog query being reset (emptied), and the expected changelog content is shown for the new project (no changelog entries).

The fix is to reset the query when switching projects, as that will give the correct behaviour if the new project is new and has no changelog entries.

That the contents of a current changelog query is kept when a new changelog fetch fails with a 404 Not found error should be treated as a bug. The API says that there isn't a changelog, so whatever data is in the GUI cache should be removed. With this behaviour no incorrect changelog would be shown when switching to a new project. Also, I noticed that when a project is initialised a changelog entry is not created - this should be done.

joargr: I've now created issues for dealing with the bugs described:

- `fmu-settings-api`: [Add changelog entry when initialising a project](https://github.com/equinor/fmu-settings-api/issues/384)
- `fmu-settings-gui`: [Reset changelog query when fetch returns 404 Not found](https://github.com/equinor/fmu-settings-gui/issues/415)

The fix of using `resetQueries` when switching projects should still be done in this PR. The new issues will in general make the code more robust.

frontend/src/components/home/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+    rms: "RMS project",
+    "rms.coordinate_system": "RMS coordinate system",
+    "rms.horizons": "RMS horizons",
+    "rms.path": "RMS project path",
+    "rms.version": "RMS version",
+    "rms.wells": "RMS wells",
+    "rms.zones": "RMS stratigraphic zones",
+  },
+  "mappings.json": {
+    stratigraphy: "stratigraphy",
```
GibranAlfa, line 35: also add wellbore in here, as seen from the video i sent for stale changelog, the entry for wellbore is different than the stratigraphy mapping

frontend/src/components/project/overview/ProjectSelector.tsx
GibranAlfa: @nourinmohd only change to resetqueries in project selector, others remain invalidatequeries

nourinmohd: Have tested it using `resetQueries`, works the way it should when switching between projects.

joargr (approved):
This looks good!

PR #354, opened 2026-04-13
MAINT: Move SMDA health check info to separate component

### LINKED ISSUE #353: Have SMDA check be a separate component
The masterdata page requires access to the SMDA API, so there is a check to make sure the necessary data is present. Other pages also require access to the SMDA API, so the SMDA check should be refactored to be a separate component for easy use in multiple pages.

joargr:
Resolves #353 

Moves the information from the SMDA health check from the masterdata page to a separate component, for easy use on any page needing SMDA access. Some renaming of variables and components is done. The new component is given the result from the SMDA health check and shows the warning box if the check fails. When the check is successful, no content is displayed. This simplifies the logic on the page requiring the check. Note that the SMDA health check itself is to be called from the page, as the result typically will be needed for other components as well.

## Checklist


GibranAlfa (approved):
looks good ✅

PR #371, opened 2026-04-29
ENH: Stratigraphy mapping of zones

### LINKED ISSUE #241: Add stratigraphy mapping/matching page/dialog

joargr:
Resolves #241 

Page for showing and editing stratigraphy mapping, for mapping RMS names to SMDA names. This PR implements mapping of zones, with mapping of horizons to be implemented in a separate issue. The data model for mappings is currently being updated, but this PR implements the data model before the update, as the API currently doesn't support the updated model. The newest version of `fmu-datamodels` implements the updated model. When running the code in this PR a previous version of `fmu-datamodels` needs to be used, a version prior to commit `38ad25cc4e0f94a94f7fb3b8d1252dfa12f85abf` from 2026-04-29 15:27.

- Similar to the masterdata page, the stratigraphy mapping page shows data with no need to SMDA access. Editing will need to be enabled, which will check for SMDA requirements
- Enables setting an SMDA name for an RMS zone, and for setting aliases for an RMS zone
- Supports displaying and editing unmappable zones. As the API currently doesn't support the new relation type `unmappable`, the mapping entry for a zone set to be unmappable uses a temporary value, which satisfies the model currently expected by the API
- A service has been created for fetching mappings. This handles 404 Not found errors gracefully
- A new common form field component is introduced: `ArrayTextField`. This is for use in editing an array value, such as an array of alias names. Note that when using this component, the code for the field itself needs to reside in the original form component, and loop through the array values. These sub fields can then be handled by the new component. The new component offers an action for removing a value, and an action for adding a new value to the array
- There is some error display: Some errors for the selection of SMDA names are shown. Errors from the API are shown as toasts
-  Fixes a bug so that wide content in the stratigraphy table gets a horizontal scrollbar

**Update 2026-05-08:**
The code has been updated so that this PR now implements the updated data model and API endpoints.

## Checklist


frontend/src/components/project/stratigraphy/functions.ts
```diff
@@ -0,0 +1,169 @@
...
+    return updated as ZoneMappings;
+  } else if (value.smdaUuid === specialOptions.unmappableZone.value) {
+    return {
+      ...zoneMappings,
+      [value.rmsName]: {
+        ...value,
+        unmappable: true,
+      },
+    };
+  } else if (value.smdaUuid === specialOptions.divider.value) {
```
joargr, line 142: There is validation in the edit form, so this value will not pass through when saving the form. Still, such a value is handled here, and results in the form value being ignored.

frontend/src/components/project/stratigraphy/functions.ts
```diff
@@ -0,0 +1,169 @@
...
+      result.push({
+        mapping_type: mappingType,
+        source_system: sourceSystem,
+        target_system: targetSystem,
+        relation_type: "primary",
+        source_id: mapping.rmsName,
+        target_id: mapping.unmappable ? tempUnmappable.id : mapping.smdaName,
+        target_uuid: mapping.unmappable
+          ? tempUnmappable.uuid
+          : mapping.smdaUuid,
```
joargr: As a temporary solution, an unmappable entry is converted to special values in the target fields.

frontend/src/components/project/stratigraphy/Overview.tsx
```diff
@@ -0,0 +1,558 @@
...
+  mutationCallback: (props: MutationCallbackProps<ZoneMapping>) => void;
+  mutationIsPending: boolean;
+  isOpen: boolean;
+  closeDialog: () => void;
+}) {
+  const form = useAppForm({
+    defaultValues: {
+      ...zoneMapping,
+      ...(zoneMapping?.smdaUuid === tempUnmappable.uuid && {
+        smdaUuid: specialOptions.unmappableZone.value,
```
joargr: Special handling for dealing with the "unmappable" option in the SMDA selection list.

frontend/src/components/project/stratigraphy/Overview.tsx
```diff
@@ -0,0 +1,558 @@
...
+          formSubmitCallback({ message: data.message, formReset });
+          closeEditDialog();
+        },
+      },
+    );
+  };
+
+  return (
+    <>
+      {editDialogOpen && (
```
joargr, line 406: The `Edit` component is rendered only when it is meant to be shown (the edit dialog is open). This ensures that the data (`activeZoneMapping`) is always defined.

frontend/src/components/project/stratigraphy/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+export function validateSelectValue(value: string) {
+  return value === specialOptions.divider.value
+    ? "A value needs to be selected"
+    : undefined;
+}
+
+export const tempUnmappable: { id: string; uuid: string } = {
+  id: "_unmappable",
+  uuid: "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee",
+};
```
joargr: These are the values stored to the API for an unmappable entry, pending the API's support of the updated mapping model.

frontend/eslint.config.js
```diff
@@ -53,7 +53,12 @@ export default tseslint.config(
         { allowConstantExport: true },
       ],
       ...reactX.configs["recommended-typescript"].rules,
-      "@typescript-eslint/no-unused-vars": "error",
+      "@typescript-eslint/no-unused-vars": [
+        "error",
+        {
+          ignoreRestSiblings: true,
```
joargr, line 59: An update to the linting rules to avoid a warning on code like this:

```ts
const { [value.rmsName]: _, ...updated } = zoneMappings;
```

Without the rule change a warning for unused variable would be shown for `_`.

frontend/src/components/project/stratigraphy/Overview.tsx
```diff
@@ -0,0 +1,558 @@
...
+          ? {
+              unmappable: lookup[key].target_uuid === tempUnmappable.uuid,
+              smdaName:
+                lookup[key].target_uuid === tempUnmappable.uuid
+                  ? specialOptions.unmappableZone.label
+                  : lookup[key].official_name,
+              smdaUuid:
+                lookup[key].target_uuid === tempUnmappable.uuid
+                  ? specialOptions.unmappableZone.value
+                  : (lookup[key].target_uuid ?? ""),
```
joargr: Special handling for dealing with an unmappable entry fetched from the API.

frontend/src/services/project.ts
```diff
@@ -112,3 +127,45 @@ export function useProject(options?: Options<ProjectGetProjectData>) {
...
+            errorStatus = error.status;
+            // Use normal handling for unauthorized response
+            if (error.status === HTTP_STATUS_UNAUTHORIZED) {
+              return Promise.reject(error);
+            }
+            if (error.response?.data && "detail" in error.response.data) {
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+              text = String(error.response.data.detail);
+            }
+          }
```
GibranAlfa: here we only rejects 401 error, but swallow any other errors as status: false and create an empty mapping. Should we only treat 404 as empty mappings? For ex if it's throwing 422 because the mappings file on disk is corrupted/invalid because it's on old mapping model, this will silently overwrite it as an empty mappings, so the user will start from new.

GibranAlfa (commented):
have tested this by matching my drogon strat with a random field i use in masterdata and it works great! we will have a discussion on Monday about the mappings model again, so this might have to be changed a bit before merging. there's also a small comment to consider

frontend/src/services/project.ts
joargr: Good point. The main objective of this service is to handle a 404 Not found error as missing mappings - an empty mappings list, instead of showing a regular error. The service can be adjusted to use normal handling for all errors except 404 Not found, however, I think it would be better if the API endpoint would reply with an empty mappings list when the mappings file doesn't exist. This will also be the same behaviour as when storing mappings. If this change is made then there would not need to be a service for getting mappings.

GibranAlfa (dismissed):
have tried matching zones with unmappable case, mappable, and mappable with alias and had no problems. I think this is good to be merged!

frontend/src/services/tasks.ts
```diff
@@ -16,38 +13,9 @@ export type Task = {
...
-          ? error
-          : new Error("Error getting mappings");
-      }
-    },
-    enabled: !!projectPath,
-    retry: false,
-  });
+  const { data: mappings } = useSuspenseQuery(
+    projectGetMappingsOptions({ path: mappingsPaths.stratigraphyRms }),
+  );
```
GibranAlfa: we fetch mappings before checking whether a project is actually selected. So even though this hook returns `[]` when there is no project, the request has already happened by then, and get mappings endpoint then return 401 unauthorized because we dont have project session yet.

Concrete example: open the app fresh, before selecting any project. The header still renders, the header uses this checklist hook, and this hook immediately asks for stratigraphy mappings. But there is no active project yet, so that request does not really make sense on that screen. Before this change, that fetch was skipped until a project existed.

I think the query should only run after the project check passes, otherwise the app is doing project-specific work in the no-project state.

I thought it was a bug locally on my api, so I ended up creating a session and selecting project manually via api endpoints and then open the gui. We don't encounter this problem if user run `fmu settings` from a project root as it automatically picked up that project. However, I think this should be addressed first.

GibranAlfa: maybe use normal query here instead of suspense query?

joargr: Good catch. The code has now been updated so that mappings are fetched only when there is an active project. This is done both on the stratigraphy mapping page as well as for the task list, used in the header and on the home page.

frontend/src/components/project/stratigraphy/functions.ts
```diff
@@ -0,0 +1,195 @@
...
+}
+
+export function createMutationValue(zoneMappings: ZoneMappings) {
+  const mappingType: MappingType = "stratigraphy";
+  const sourceSystem: DataSystem = "rms";
+  const targetSystem: DataSystem = "smda";
+  const result: InternalStratigraphyMappingsOutput = [];
+
+  Object.values(zoneMappings).forEach((mapping) => {
+    if (mapping.smdaUuid !== "" || mapping.unmappable) {
```
GibranAlfa: This currently only serializes zones that have an SMDA UUID or are explicitly marked unmappable. That means a zone left unmapped (SMDA value of (not set)) is not saved at all, and alias-only edits for an unmapped zone are dropped from the payload. I think we want to save the not yet mapped case as rms->rms primary and add rms->rms alias if it exists in the mappings.json.

GibranAlfa (approved):
The bug has been fixed, and I have a small comment on the not-mapped-yet case, but it's not a blocker for this to be merged.

tnatt (approved):
Have not looked at the code, just tested the functionality.
Tried to break it, but seems to be quite robust and it works mainly as expected! Nice job 🙂 
- main issue I discovered was that if you edit a zone that have aliases and you set the smda name to `(not set)` and stores it the aliases is removed.

I agree with @GibranAlfa that it can be merged, and I'll write my other improvement potential suggestions in a separate issue.

frontend/src/components/project/stratigraphy/functions.ts
joargr: This has now been corrected, so that a zone with only aliases will be stored as expected.

PR #388, opened 2026-05-18
BUG: Retry SMDA requests after refreshing SSO token

### LINKED ISSUE #346: Failed SMDA requests due to stale token are not auto-retried
After PR #275, the frontend still refreshes a stale SMDA SSO access token, but it no longer retries the original failed SMDA request.

## Current behavior

If an SMDA request fails with `401` because the SSO access token is stale:

- the frontend refreshes the token
- the refreshed token is added to the backend session
- the original request still fails
- the user must repeat the action manually

## Expected behavior

For normal SMDA requests, the frontend should:

- silently refresh the stale SSO access token
- update the backend session
- retry the original failed request once

The user should not need to search again manually.

This should **not** apply to failed `smda/health` requests. That exclusion added in #275 should stay.

A fix should:

- retry only once
- avoid infinite retry loops
- keep the `smda/health` exception
- update the backend session token before retrying

GibranAlfa:
Resolves #346 

Fixes stale SMDA SSO token requests by refreshing the token, patching it into the backend session, then retrying the failed request once.

I also moved the token refresh + session patching into one shared helper in `main.tsx`, so we keep the #275 idea of handling SSO token logic centrally. This avoids patching the same SSO token multiple times because the `ACQUIRE_TOKEN_SUCCESS` handler now only updates local state, while the shared helper handles the backend session patch.

Have tested this by changing the token manually and call the field search in masterdata again, it refreshed the token and retried the failed request automatically.

## Checklist

### DIFF
--- a/frontend/src/main.tsx
+++ b/frontend/src/main.tsx
@@ -3,6 +3,7 @@ import {
   type EventMessage,
   EventType,
   InteractionRequiredAuthError,
+  InteractionType,
   PublicClientApplication,
 } from "@azure/msal-browser";
 import { MsalProvider, useMsal } from "@azure/msal-react";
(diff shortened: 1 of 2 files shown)


nourinmohd (commented):
Gave this a quick go, works well 👍 .

joargr:
I've only looked at this change very briefly, but how it does it handle parallel API calls to SMDA endpoints? If these respond with an error, will the new code lead to acquiring and patching SSO access token happening multiple times in parallel? This should be avoided, as acquiring and patching should happen only once. A situation where such parallel API calls happen is when getting masterdata from multiple fields in the masterdata editing.

GibranAlfa:
> I've only looked at this change very briefly, but how it does it handle parallel API calls to SMDA endpoints? If these respond with an error, will the new code lead to acquiring and patching SSO access token happening multiple times in parallel? This should be avoided, as acquiring and patching should happen only once. A situation where such parallel API calls happen is when getting masterdata from multiple fields in the masterdata editing.

good catch! before the added guard, the GUI still worked fine, but they acquire and patch SSO token multiple times. Now with the added guard, it will only acquire and patch once.

joargr (approved):
I haven't tried to trigger the situation, but it seems that the updated code can deal with the situation successfully, with the necessary retries.

GibranAlfa:
small fix: also patch the SMDA token if interactive login is required, without bringing back duplicate patches for silent refreshes. Also move MSAL event callback cleanup to the useEffect cleanup so old callbacks are removed correctly.

PR #389, opened 2026-05-18
ENH: Stratigraphy mapping of horizons

### LINKED ISSUE #372: Add mapping of stratigraphic horizons
Add mapping of horizons to the existing stratigraphic mapping page, which now maps zones.

joargr:
Resolves #372 

Implements display and editing of stratigraphy mapping of horizons.

- Horizons are now displayed similar to zones, with a horizon shown as a box. The RMS name and the SMDA name are shown in the box. The background colour of the horizon box is white, with the zone box being green
- The CSS grid layout is changed to better show that the stratigraphy grid lines and the horizons exist between zones. The grid lines now touch the horizons
- The horizon box is shown with the same border line style as the stratigraphy grid line, either solid or dashed, indicating an interpreted or calculated horizon
- The updated layout of the stratigraphic framework also applies to the RMS page, as both this and the stratigraphy mapping page use the same basic layout code
- As the horizon and zone mostly share the same display, it is now the same component that is used to show both horizons and zones. This makes for easier maintenance of the code
- Functions and variables have been renamed to align with shared usage for horizons and zones
- Added a tip explaining that the options when choosing SMDA names for horizons are based on the already selected SMDA names for zones, so that zones should be edited before horizons

## Checklist


frontend/src/components/project/stratigraphy/Overview.tsx
```diff
@@ -89,29 +95,31 @@ const { useAppForm } = createFormHook({
...
   isOpen: boolean;
   closeDialog: () => void;
 }) {
   const form = useAppForm({
     defaultValues: {
-      ...zoneMapping,
-      ...(zoneMapping?.unmappable && {
+      ...elementMapping,
+      ...(elementMapping?.unmappable && {
         smdaUuid: specialOptions.unmappableZone.value,
```
GibranAlfa: This should choose `unmappableHorizon` or `unmappableZone` based on `elementMapping.elementType`. For an unmappable horizon, this currently sets `_unmappableZone`, but the horizon select options contain `_unmappableHorizon`, so reopening an unmappable horizon gives the form a value that is not in its option list.

frontend/src/components/project/stratigraphy/Overview.tsx
```diff
@@ -127,21 +135,24 @@ function Edit({
...
       form.setFieldMeta,
       "smdaUuid",
-      smdaNameOptions,
+      elementMapping?.elementType
+        ? smdaNameOptions[elementMapping.elementType]
+        : [],
       {
-        value: zoneMapping?.unmappable
+        value: elementMapping?.unmappable
           ? specialOptions.unmappableZone.value
```
GibranAlfa: same as above

frontend/src/components/project/stratigraphy/Overview.tsx
```diff
@@ -531,15 +673,26 @@ export function Overview({
...
+          {editMode &&
+            !projectReadOnly &&
+            smdaHealthStatus &&
+            (stratigraphicColumn ? (
+              <PageText>
+                💡 When editing horizons the options for SMDA names are drawn
+                from the top and base horizons of the zones that have SMDA
+                names. SMDA names should thus be set for zones first, so that
+                the SMDA names for horizons are available for the horizon
+                editing
```
GibranAlfa: I think we can make this clearer, something like “Set SMDA names for zones first. Horizon options are derived from the top and base horizons of mapped zones.”

frontend/src/components/project/stratigraphicFramework/functions.ts
```diff
@@ -15,16 +15,29 @@ export function useFrameworkData() {
...
   const gridPlacement = new Map<string, ZonePlacementInfo>();
 
   zones.forEach((zone) => {
     const topHorizonIndex = findIndexByName(horizons, zone.top_horizon_name);
     const baseHorizonIndex = findIndexByName(horizons, zone.base_horizon_name);
+    if (topHorizonIndex < 0 || baseHorizonIndex < 0) {
+      console.warn(
+        "Top or base horizon for zone",
+        zone.name,
+        "not found when calculcating grid placement, ignoring this zone",
```
GibranAlfa: typo: *calculating

frontend/src/components/project/stratigraphicFramework/functions.ts
```diff
@@ -15,16 +15,29 @@ export function useFrameworkData() {
   return data;
 }
 
+export function getHorizonLineStyle(horizon: RmsHorizon): HorizonLineStyle {
+  return horizon.type.startsWith("interpreted") ? "solid" : "dashed";
```
GibranAlfa, line 19: i'm not sure if the users are familiar with solid line being interpreted and dashed line being calculated. if they are not familiar, maybe add a tooltip or hint explaining this?

GibranAlfa (commented):
looks good overall! just one bug that if we set an unmappable horizon, and then we open the dialog again, it does not show the unmappable value properly because it uses `_unmappableZone` as the value, which is not recognized in horizon option list.

frontend/src/components/project/stratigraphy/Overview.tsx
joargr: Good catch. Indeed, this handling needed to be updated. This is now fixed.

frontend/src/components/project/stratigraphy/Overview.tsx
joargr: Fixed.

frontend/src/components/project/stratigraphy/Overview.tsx
joargr: I agree, this is a better and more concise description.

frontend/src/components/project/stratigraphicFramework/functions.ts
joargr: Fixed.

frontend/src/components/project/stratigraphicFramework/functions.ts
joargr: On the RMS page the tooltip shown when hovering over a horizon gives the type: interpreted or calculated. As the layout and presentation of horizons and zones otherwise is quite similar between the RMS page and the stratigraphy mapping page I think it's enough to give this information only on the RMS page. There are also tooltips for the zones on the RMS page, so the general mode is that such extra information is given on the RMS page, leaving the stratigraphy mapping page open for showing multiple names and an edit action for each horizon and zone. And when setting up RMS information for the project, both these pages will be visited by the user so the user will get information from both places.

GibranAlfa (approved):
looks good to me! all concerns are addressed and i don't see any other blockers, should be good to merge.

PR #390, opened 2026-05-19
BUG:Pressing ESC closes topmost dialog

### LINKED ISSUE #360: Pressing Esc on masterdata field search closes the edit dialog
Having the masterdata edit dialog open and then opening the field search dialog, when pressing the Esc key it is the masterdata editing dialog that closes, not the field search dialog. This is a bug, and is a confusing behaviour.

Currently the field search dialog is not set as dismissible, ie. that pressing the Esc key can close it. I think it's ok that this dialog is dismissible, with a confirm dialog for closing it if there are changes (ie. that a search has been performed and the results are shown). Perhaps defining the field search dialog as dismissible will result in correct handling of the Esc key press (closing the field search dialog and not the masterdata editing). Otherwise this needs to be handled so that the Esc key only closes the uppermost dialog, as is expected.

nourinmohd:
Resolves #360 

Possible fix to close the field search dialog first then the masterdata field dialog. 

## Checklist

### DIFF
--- a/frontend/src/components/project/masterdata/FieldSearch.tsx
+++ b/frontend/src/components/project/masterdata/FieldSearch.tsx
@@ -9,7 +9,6 @@ import { type Dispatch, type SetStateAction, useEffect, useState } from "react";
 
 import type { SmdaFieldSearchResult, SmdaFieldUuid } from "#client";
 import { smdaPostFieldOptions } from "#client/@tanstack/react-query.gen";
-import { ConfirmCloseDialog } from "#components/common";
 import { CancelButton, GeneralButton } from "#components/form/button";
 import { SearchFieldForm } from "#components/form/form";
 import { EditDialog, PageSectionSpacer, PageText } from "#styles/common";
@@ -107,7 +106,6 @@ export function FieldSearch({
 }) {
   const [searchValue, setSearchValue] = useState("");
   const [selectedFields, setSelectedFields] = useState<Array<string>>([]);
-  const [confirmCloseDialogOpen, setConfirmCloseDialogOpen] = useState(false);
 
   const { data } = useQuery({
     ...smdaPostFieldOptions({ body: { identifier: searchValue } }),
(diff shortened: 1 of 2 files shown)


frontend/src/components/project/masterdata/FieldSearch.tsx
```diff
@@ -114,41 +116,84 @@ export function FieldSearch({
...
-    </EditDialog>
+          <CancelButton onClick={handleCloseRequest} />
+        </Dialog.Actions>
+      </EditDialog>
+
+      <ConfirmCloseDialog
+        isOpen={confirmCloseDialogOpen}
+        handleConfirmCloseDecision={handleConfirmCloseDecision}
+        title="Discard search results"
+        description="You have selected fields that haven't been added to the masterdata yet."
```
GibranAlfa: consider using "You have selected fields that have not yet been added to the masterdata."

GibranAlfa (approved):
I have tested this and it fixed the bug 👍 very nice

frontend/src/components/project/masterdata/FieldSearch.tsx
```diff
@@ -114,41 +116,84 @@ export function FieldSearch({
...
-          }}
-        />
-        <CancelButton onClick={handleClose} />
-      </Dialog.Actions>
-    </EditDialog>
+          <CancelButton onClick={handleCloseRequest} />
+        </Dialog.Actions>
+      </EditDialog>
+
+      <ConfirmCloseDialog
```
joargr: I think it's a bit of an overkill to have a confirm close dialog on the field search. Should the user have performed a search and selected one or more fields, and then mistakenly close the search dialog without adding the selected fields to the masterdata editing, there wouldn't be much effort required to recreate the search results and field selections. Other dialog forms, like the masterdata editing or the RMS horizon and zone selection, are forms that the user is expected to spend more time and effort editing, so a mistakenly closed dialog here would have more of a negative effect.

I think it is more likely that when a user closes the field search dialog, typically by using the Cancel button, it is because the user wants to abort the addition of a field. Then the confirm close dialog will simply be an extra, slightly annoying step when accomplishing what the user already has decided.

GibranAlfa: to be fair nourin is just following the issue description here, but yes it's a bit overkill here and we can just remove it

nourinmohd: Yes I did follow the issue description but I also found adding the confirm dialog quite an addition. I'll remove it

joargr: You're right, the issue does suggest having a confirm close dialog on the search results. Seeing how it looks in practice I think it's better to not have that, and just close the search dialog it immediately.

frontend/src/components/project/masterdata/FieldSearch.tsx
```diff
@@ -114,41 +116,84 @@ export function FieldSearch({
 
   function handleClose() {
     setSearchValue("");
+    setSelectedFields([]);
```
joargr: `selectedFields` is set based on the contents of `selectedRows`, and `selectedRows` is reset on every search. There shouldn't be a need of explicitly resetting `selectedFields`, so if the functionality for the confirm close dialog is removed then the reset of `selectedFields` should also be removed, keeping the original function unchanged.

joargr (approved):
Nice!

PR #391, opened 2026-05-19
ENH: Add link to Documentation

### LINKED ISSUE #383: Add link to FMU Settings documentation
From FMU Settings GUI we should have a link to the FMU Settings documentation page. 

This should be done in the same way as Sumo has done it.

krhart:
Resolves #383 

Added link to FMU Settings documentation to Header

## Checklist

### DIFF
--- a/frontend/src/components/Header.tsx
+++ b/frontend/src/components/Header.tsx
@@ -12,6 +12,7 @@ import {
   check,
   checkbox,
   comment,
+  external_link,
   warning_filled,
 } from "@equinor/eds-icons";
 import { tokens } from "@equinor/eds-tokens";
@@ -279,7 +280,19 @@ export function Header() {
         </TopBar.Header>
         <TopBar.Actions>
           <TaskIndicator />
+
           <FeedbackDialog />
+
+          <HeaderActionButton
+            variant="ghost"
+            forwardedAs="a"
+            href="https://equinor.github.io/fmu-settings/"
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            Documentation <Icon data={external_link} size={18} />
+          </HeaderActionButton>
+
           <ProjectInfo />
         </TopBar.Actions>
       </TopBar>


frontend/src/components/Header.tsx
```diff
@@ -111,6 +112,16 @@ function ProjectInfo() {
   );
 }
 
+function Documentation() {
+  const url = "https://equinor.github.io/fmu-settings/index.html#";
```
joargr: The url should be simplified to `https://equinor.github.io/fmu-settings/`.

frontend/src/components/Header.tsx
```diff
@@ -111,6 +112,16 @@ function ProjectInfo() {
...
+
+  return (
+    <HeaderActionButton onClick={() => window.open(url, "_blank")}>
+      Documentation <Icon data={external_link} size={18} />
+    </HeaderActionButton>
+  );
+}
```
GibranAlfa: to follow the existing codebase pattern, we should do it like this
```suggestion
function Documentation() {
  const url = "https://equinor.github.io/fmu-settings/";

  return (
    <HeaderActionButton
      forwardedAs="a"
      href={url}
      target="_blank"
      rel="noopener noreferrer"
    >
      Documentation <Icon data={external_link} size={18} />
    </HeaderActionButton>
  );
}
```

krhart: Then it will look weird on the Header, like a ordinary link (underscore, blue color etc), not like the other buttons on the Header. Should I then just add styling to the `a` ?
image

GibranAlfa: i edited my comment, it should use `forwardedAs` instead of `as`

joargr: I agree that a regular `Button` should be used, with a traditional `href`. That will simplify the code and it will look the same, and in fact be implemented like the [documentation link](https://github.com/equinor/sumo-ui/blob/a053a58b50db0f38fe2696e4e06c5c3ac0b9ff25/src/components/shared/Header/Header.jsx#L123-L136) in Sumo. Suggest doing it as follows:

```html
  <Button
    forwardedAs="a"
    href="https://equinor.github.io/fmu-settings/"
    target="_blank"
    rel="noopener noreferrer"
  >
    Documentation <Icon data={external_link} size={18} />
  </Button>
```

Edited: To keep the background the same, the `HeaderActionButton` should probably be used instead of a regular `Button`. So @GibranAlfa code should do the trick, but I would suggest just putting the URL directly into the `href` property instead of setting it in a variable.

Also, as the `Documentation` function just returns the button code, I think the button code can just be put directly into `TopBar.Actions`: 

```html
  <TopBar.Actions>
    <TaskIndicator />

    <FeedbackDialog />

    <HeaderActionButton
      variant="ghost"
      as="a"
      href="https://equinor.github.io/fmu-settings/"
      target="_blank"
      rel="noopener noreferrer"
    >
      Documentation <Icon data={external_link} size={18} />
    </HeaderActionButton>

    <ProjectInfo />
  </TopBar.Actions>
```

Note the empty lines between each action, that should make the code a bit easier to read.

GibranAlfa: I think we should use HeaderActionButton like what we use in the feedback dialog

GibranAlfa (approved):
Looks good!

joargr (approved):
Nice!

PR #394, opened 2026-05-20
BUG: Use form.subscribe to set submit button disable

### LINKED ISSUE #83: Use `form.Subscribe` to set submit button disable state
The `TextField` component from `#components/form/field` has the functionality to edit this disabled state of the form's submit button if the field is not valid or the value is the default value. 

If there are multiple Textfield components in a form this can trigger a bug:
1. Edit in field one - submit is `Enabled`
2. Edit in field two - submit is still `Enabled`
3. Remove the edits in field two back to default value - submit is `Disabled` even though field one still have edits 

The solution is to remove the option to set the disabled state for a submit inside the `TextField` component, and rather set it from the form state directly using `form.Subscribe`. The `form.state.isDefaultValue` can be used to check if all values for the entire form are at their default values. The `form.state.canSubmit` will check if any field is invalid (set from their validators).

```typescript
            <form.Subscribe>
              {(state) => (
                <form.SubmitButton
                  label="Save"
                  disabled={state.isDefaultValue || !state.canSubmit}
                  isPending={isPending}
                />
              )}
            </form.Subscribe>
```

nourinmohd:
Resolves #83 

Use form.subscribe to set submit button disable

## Checklist

### DIFF
--- a/frontend/src/components/form/form.tsx
+++ b/frontend/src/components/form/form.tsx
@@ -121,11 +121,13 @@ export function EditableTextFieldForm({
             />
           ) : (
             <>
-              <form.Subscribe>
-                {(state) => (
+              <form.Subscribe
+                selector={(state) => [state.isDefaultValue, state.canSubmit]}
+              >
+                {([isDefaultValue, canSubmit]) => (
                   <form.SubmitButton
                     label="Save"
-                    disabled={state.isDefaultValue || !state.canSubmit}
+                    disabled={isDefaultValue || !canSubmit}
                     isPending={mutationIsPending}
                     helperTextDisabled="Value can be submitted when it has been changed and is valid"
                   />
--- a/frontend/src/components/form/form.tsx
+++ b/frontend/src/components/form/form.tsx
@@ -59,7 +59,6 @@ export function EditableTextFieldForm({
   mutationIsPending,
 }: EditableTextFieldFormProps) {
   const [isReadonly, setIsReadonly] = useState(true);
-  const [submitDisabled, setSubmitDisabled] = useState(true);
 
   const validator = handleValidator({ length, minLength, initialValue: value });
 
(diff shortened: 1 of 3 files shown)


nourinmohd:
@tnatt  I took a jab at this because the issue was calling out to me in my dreams, review it whenever you can there is no rush. Same goes to you guys @GibranAlfa @joargr

GibranAlfa (approved):
Looks good to me, although I can't find a concrete way to reproduce the bug, it seems like this can only happen in project selector, but we open the project right away, so the disabled button when we select a project doesn't matter.

That said, removing submit button state handling from `TextField` still makes sense. The field component shouldn’t be responsible for deciding whether the whole form can submit, and using `form.Subscribe` with form-level state is the safer pattern.

Also, this is a bug fix, not maintenance, so use BUG prefix, not everything is MAINT.

frontend/src/components/form/form.tsx
```diff
@@ -123,12 +121,16 @@ export function EditableTextFieldForm({
...
+              <form.Subscribe>
+                {(state) => (
+                  <form.SubmitButton
+                    label="Save"
+                    disabled={state.isDefaultValue || !state.canSubmit}
+                    isPending={mutationIsPending}
+                    helperTextDisabled="Value can be submitted when it has been changed and is valid"
+                  />
+                )}
+              </form.Subscribe>
```
joargr: It is preferable to use a selector with the subscribe, as it allows for narrowing which part of the state will be used, and thus minimizes the number of times rerenders happen.

```suggestion
              <form.Subscribe
                selector={(state) => [state.isDefaultValue, state.canSubmit]}
              >
                {([isDefaultValue, canSubmit]) => (
                  <form.SubmitButton
                    label="Save"
                    disabled={isDefaultValue || !canSubmit}
                    isPending={mutationIsPending}
                    helperTextDisabled="Value can be submitted when it has been changed and is valid"
                  />
                )}
              </form.Subscribe>
```

nourinmohd: Implemented. Switched to a selector-based subscribe so this component only rerenders on changes to isDefaultValue/canSubmit.

joargr (approved):
Nice!

PR #402, opened 2026-05-27
DEV: Set Dependabot to Ignore major version updates to EDS

### LINKED ISSUE #401: Set Dependabot to ignore major version updates for EDS packages
Currently Dependabot will propose updating all npm packages in `fmu-settings-gui` to the newest versions, regardless of whether it is a major, minor or patch version changes. EDS (Equinor Design System) is currently used at major version 0, when the current stable version is 1 and version 2 is in beta. Upgrading to newer major versions should be handled manually, as it will require a significant refactoring of code.

Dependabot should be configured to not propose updating to newer major versions. This setting could possible be a permanent change, as updating to newer major versions would most likely require manual checking and refactoring.

joargr:
Resolves #401

Configures Dependabot to ignore major version updates to EDS (Equinor Design System), and thus not create PRs with such update proposals.

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,6 +26,9 @@ updates:
       interval: "weekly"
     allow:
       - dependency-type: "all"
+    ignore:
+      - dependency-name: "@equinor/eds-*"
+      - update-types: ["version-update:semver-major"]
     groups:
       all-deps:
         patterns:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -28,7 +28,7 @@ updates:
       - dependency-type: "all"
     ignore:
       - dependency-name: "@equinor/eds-*"
-      - update-types: ["version-update:semver-major"]
+        update-types: ["version-update:semver-major"]
     groups:
       all-deps:
         patterns:


GibranAlfa (approved):

.github/dependabot.yml
```diff
@@ -26,6 +26,9 @@ updates:
       interval: "weekly"
     allow:
       - dependency-type: "all"
+    ignore:
+      - dependency-name: "@equinor/eds-*"
+      - update-types: ["version-update:semver-major"]
```
GibranAlfa: i think this will ignore all eds package, to only ignore semver, we should do
```suggestion
    ignore:
      - dependency-name: "@equinor/eds-*"
        update-types: ["version-update:semver-major"]
```
ref https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/controlling-dependencies-updated#specifying-the-semantic-versioning-level-to-ignore

GibranAlfa (changes requested):

.github/dependabot.yml
joargr: Yes, of course, good catch!

GibranAlfa (approved):

PR #407, opened 2026-05-28
ENH: Update definition of done in stratigraphy task lists

### LINKED ISSUE #385: Task list: Fix definition of done of stratigraphy mappings
Fix definition of done of stratigraphy mappings

Definition of done: A mapping decision has been made (selected a name or set unmappable) for every Zone and Horizon from the project config.

nourinmohd:
Resolves #385 

Fixed definition of done for Stratigraphy mapping in tasks list. 

## Checklist

### DIFF
--- a/frontend/src/services/tasks.ts
+++ b/frontend/src/services/tasks.ts
@@ -23,6 +23,13 @@ export function useTaskList(): Task[] {
   }
 
   const config = project.data.config;
+  const zones = config.rms?.zones ?? [];
+  const horizons = config.rms?.horizons ?? [];
+  const mappedRmsIds = new Set(
+    (mappings?.stratigraphy ?? [])
+      .filter((m) => m.source_system === "rms")
+      .map((m) => m.source_id),
+  );
 
   return [
     {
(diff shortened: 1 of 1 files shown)


frontend/src/services/tasks.ts
```diff
@@ -23,6 +23,13 @@ export function useTaskList(): Task[] {
   }
 
   const config = project.data.config;
+  const zones = config.rms?.zones ?? [];
+  const horizons = config.rms?.horizons ?? [];
+  const mappedRmsIds = new Set(
+    (mappings?.stratigraphy ?? [])
+      .filter((m) => m.source_system === "rms")
```
GibranAlfa: we should check if it's mapped to smda (like joar did), not just if every rms is mapped, because we can have alias only mapping, and it will be considered done with this implementation. Definition of done should be if every rms source id is mapped, either mapped to smda or zone doesn't exist in SMDA (unmappable). Something like this
```ts
.filter(
  (m) =>
    m.source_system === "rms" &&
    m.target_system === "smda" &&
    (m.relation_type === "primary" || m.relation_type === "unmappable"),
)

frontend/src/services/tasks.ts
GibranAlfa: RMS task currently requires both zones and horizons, but I think there may be a use case that we only need horizon mapping, should confirm with @tnatt. If so, then change this && to || as well, and since you make a new zones and horizons const, update this as well

nourinmohd: Alright. I'll wait on @tnatt to confirm and make it in one commit.

GibranAlfa (approved):
LGTM, maybe wait for Joar's opinion as well

joargr (approved):
This looks good.

PR #413, opened 2026-06-01
ENH: Set specific styling for unmappable and not-set stratigraphic mappings

### LINKED ISSUE #382: Use different colors for Stratigraphic mappings that are completed
On the Stratigraphic mappings page we should add some visual feedback to make it even more clear when a mapping is set/completed and when it is not. Using different colors e.g.

### LINKED ISSUE #396: User feedback: No zone and (not set) options looks too similar
During FMU Results workshop we got feedback from some of the users that the "No zone" and (not set) status for a stratigraphic mapping looks too similar. 

We should look into making it more clear that the is a big difference here. 

Using different colors for different status might be good enough (will be solved through https://github.com/equinor/fmu-settings-gui/issues/382), but we should discuss if there is more we can do as well.

joargr:
Resolves #382
Resolves #396

Sets specific styling for mappings which are unmappable and mappings which are not set.

- An unmappable mapping has a red-ish background colour, indicating something which doesn't exist. The text "No horizon"/"No zone" is in italics, indicating that it is not an actual name of an element. As the background colour is light, the text color is dark
- A not-set mapping does not show a visible box. The system name ("SMDA") is shown, indicating the name of the system where the mapping has to be set. For future mapping implementations (such as wellbores) there might be more than one system to map to, so the presence of the system name for not-set mappings is needed. The text shown is "(not set)", which is the same text used elsewhere in the application to indicate values that are not set. The font style is in italics to indicate that the text is not an actual name of an element

Skjermbilde 2026-06-01 kl  15 35 25

Issue #382 has a comment about using colours from SMDA for mappings that are set. If implemented, such colours would probably not have been used as the main background colour of an SMDA name box, as it is more important to consistently use the same background colour (blue) for all mapped names. Rather, the SMDA colour could have been added to the SMDA name box, possibly as a stripe at the bottom. However, the SMDA colours would have needed to be stored as part of the mapping data in the FMU project, so as to be available also when the user does not have SMDA access. Storing of such additional information is not currently provided for the mappings data, so usage of such colour data is not implemented here.

## Checklist

### DIFF
--- a/frontend/src/components/project/stratigraphy/Overview.style.ts
+++ b/frontend/src/components/project/stratigraphy/Overview.style.ts
@@ -1,11 +1,41 @@
 import { tokens } from "@equinor/eds-tokens";
-import styled from "styled-components";
+import styled, { css } from "styled-components";
 
 import type {
   HorizonLineStyle,
   ZonePlacementInfo,
 } from "../stratigraphicFramework/types";
 import type { ElementType } from "./types";
+
+function getColors(
+  isTargetSystem: boolean,
+  isUnmappable: boolean,
+  isMissingValue: boolean,
+) {
+  let backgroundColor = "inherit";
+  let color = "inherit";
+
+  if (isTargetSystem) {
+    if (isUnmappable) {
+      backgroundColor = tokens.colors.infographic.substitute__pink_salmon.hex;
+      color = tokens.colors.text.static_icons__default.hex;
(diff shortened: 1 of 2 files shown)


GibranAlfa (approved):
tried with Drogon, LGTM 👍

Image

small nit, Resolves #382 and #396 will only resolve 382, need to do
Resolves #382
Resolves #396

PR #419, opened 2026-06-04
ENH: Enable using full width for stratigraphic framework

### LINKED ISSUE #387: Look into utilizing a bigger part of the screen size for mapping
Stratigraphy mappings can take up a big part of the screen. We should look into if we can utilize a bigger part of the screen when we do this. 

An alternative is to add an icon the user can click to expand the mappings table to use the full width of the page.

joargr:
Resolves #387 

Enables using full width for the stratigraphic tables on the RMS and Stratigraphy pages.

There are two parts to this implementation:

### Enabling setting width constraint on sections instead of on the page as a whole

The main layout sets a width constraint on the content part of the page, set to `55em`. This is part of `PageContainer` used in the root route (`routes/__root.tsx`). This is now updated so that if there is a child with the class name `notWidthConstrained` the width constraint will not apply:

```ts
export const PageContainer = styled.div`
  &:not(:has(> .notWidthConstrained)) {
    max-width: 55em;
  }
`;
```

On pages that has sections that will be using full width, the top level content needs to be wrapped in the new `PageContainerNotWidthConstrained` component, which will affect the width constraint rule on the regular `PageContainer` component. Then all sections of the page that will *not* be using full width needs to be wrapped in the new `PageSectionWidthConstrained` component, as this sets the width constraint. This component can wrap multiple sections, ie. `PageHeader`, `PageText` and `PageCode`.

Components that will use full width also need to be wrapped in `PageContainerNotWidthConstrained`, but this time with the `$notConstrained` property set to `true`. This leads to no width constraint being set.

### Adding expand width function to stratigraphic table

The stratigraphic table showing horizons and zones, used on the RMS and Stratigraphy page, now has an icon in the top right corner. The table is initially set to the same width as the rest of the page, but clicking on this icon expands the table to use the full width of the page. The width can then be collapsed again. Whether the table is expanded or not is kept in state (using session storage), so the user can navigate to other pages and come back to the same table width as before.

Expansion and collapsing of the table width is handled by toggling the `$notConstrained` property of the `PageContainerNotWidthConstrained` component that the table is wrapped in. Note that whether the stratigraphic framework should offer the option to expand the width is controlled by the property `enableWidthExpansion`. The table is also used in an edit dialog, so there this option is not set.

## Checklist

### DIFF
--- a/frontend/src/components/project/stratigraphicFramework/StratigraphicFramework.style.ts
+++ b/frontend/src/components/project/stratigraphicFramework/StratigraphicFramework.style.ts
@@ -29,9 +29,17 @@ export const StratigraphicFrameworkHeader = styled.div.attrs<{
   div {
     padding: ${tokens.spacings.comfortable.x_small} ${tokens.spacings.comfortable.small};
 
-    &:nth-child(2) {    
+    &:nth-child(2) {
       display: flex;
       justify-content: center;
+      grid-row: 1;
+      grid-column: 2 / -1;
+    }
+
+    &:nth-child(3) {
+      display: flex;
+      justify-content: right;
+      grid-row: 1;
       grid-column: 2 / -1;
     }
   }
--- a/frontend/src/components/project/stratigraphicFramework/StratigraphicFramework.style.ts
+++ b/frontend/src/components/project/stratigraphicFramework/StratigraphicFramework.style.ts
@@ -22,6 +22,8 @@ export const StratigraphicFrameworkHeader = styled.div.attrs<{
     gridTemplateColumns: `minmax(max-content, 2fr) repeat(${$numStratColumns}, 3fr)`,
   },
 }))`
+  padding-bottom: ${tokens.spacings.comfortable.small};
+
   display: grid;
   font-weight: ${tokens.typography.table.cell_header.fontWeight};
   font-size: ${tokens.typography.table.cell_text.fontSize};
@@ -41,6 +43,10 @@ export const StratigraphicFrameworkHeader = styled.div.attrs<{
       justify-content: right;
       grid-row: 1;
       grid-column: 2 / -1;
+
+      position: relative;
+      top: -8px;
+      height: 16px;
     }
   }
 `;
--- a/frontend/src/components/project/stratigraphicFramework/StratigraphicFramework.style.ts
+++ b/frontend/src/components/project/stratigraphicFramework/StratigraphicFramework.style.ts
@@ -32,21 +32,22 @@ export const StratigraphicFrameworkHeader = styled.div.attrs<{
     padding: ${tokens.spacings.comfortable.x_small} ${tokens.spacings.comfortable.small};
 
     &:nth-child(2) {
-      display: flex;
-      justify-content: center;
       grid-row: 1;
       grid-column: 2 / -1;
+
+      display: flex;
+      justify-content: center;
     }
 
     &:nth-child(3) {
-      display: flex;
-      justify-content: right;
       grid-row: 1;
       grid-column: 2 / -1;
-
-      position: relative;
-      top: -8px;
       height: 16px;
+
+      display: flex;
+      justify-content: right;
+      align-items: center;
+
     }
   }
 `;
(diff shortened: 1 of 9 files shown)


frontend/src/components/project/stratigraphicFramework/StratigraphicFramework.tsx
```diff
@@ -71,35 +96,63 @@ export function StratigraphicFramework({
...
-        <div>Zones</div>
-      </StratigraphicFrameworkHeader>
+        <StratigraphicFrameworkHeader
+          $numStratColumns={frameworkData.numStratColumns}
+        >
+          <div>Horizons</div>
+          <div>Zones</div>
+          {enableWidthExpansion && (
+            <div>
+              <Icon
```
GibranAlfa: is it better to wrap this inside an EDS button with ghost_icon variant? right now it's just an Icon component which happens to have onClick behavior, hovering the cursor to this doesn't change the cursor from pointer to hand cursor

frontend/src/components/project/stratigraphicFramework/StratigraphicFramework.tsx
```diff
@@ -71,35 +96,63 @@ export function StratigraphicFramework({
...
+                data={notWidthConstrained ? collapse : expand}
+                title={
+                  notWidthConstrained
+                    ? "Expand to full width"
+                    : "Collapse width"
```
GibranAlfa: this should be the other way around
```suggestion
                    ? "Collapse width"
                    : "Expand to full width"
```

frontend/src/styles/common.ts
```diff
@@ -8,8 +8,23 @@ import {
 import { tokens } from "@equinor/eds-tokens";
 import styled from "styled-components";
 
+const WidthConstraint = "55em";
```
GibranAlfa, line 11: wdyt about renaming PageContainerNotWidthConstrained and prop $notConstrained to PageContainerFullWidth and prop $fullWidth?

frontend/src/components/project/stratigraphicFramework/StratigraphicFramework.tsx
joargr: Good suggestion. The icon is now part of an EDS Button component, so it has a hover effect and the hand cursor is used.

frontend/src/components/project/stratigraphicFramework/StratigraphicFramework.tsx
joargr: Fixed.

frontend/src/styles/common.ts
joargr: The reasoning for the "notWidthConstrained" name is very basic - it simply overrides what the original `PageContainer` sets, which is a constraint on the width in the form of a `max-width` value. Without this width constraint the content will use all of the available width. This can be described as using the full width, however, I think the name should reflect only what is accomplished by using the new component. How the override is used, ie. what is set on the content, should not be implied from the name (so it's not named "fullWidth"). Perhaps the "notWidthConstrained" component is used with a child component to set a new but wider width constraint? So even though "fullWidth" could be used as the name, I think there's a subtle difference in simply using the name "notWidthConstrained".

frontend/src/components/project/stratigraphicFramework/StratigraphicFramework.style.ts
```diff
@@ -22,17 +22,31 @@ export const StratigraphicFrameworkHeader = styled.div.attrs<{
...
+
+    &:nth-child(3) {
+      display: flex;
+      justify-content: right;
+      grid-row: 1;
       grid-column: 2 / -1;
+
+      position: relative;
+      top: -8px;
+      height: 16px;
```
GibranAlfa, line 45: this will show the same visual layout, but with less manual adjustment

```suggestion
    &:nth-child(3) {
      display: flex;
      align-items: center;
      justify-content: right;
      grid-row: 1;
      grid-column: 2 / -1;

      height: 16px;
```

joargr: Yes, indeed. Better to use this than relative positioning.

GibranAlfa (approved):
Looks very good!

PR #421, opened 2026-06-04
MAINT: Small improvement of persisted value handling

### LINKED ISSUE #420: Update session storage in parallel with React state
For values stored in React state as well as in the browser's session storage, a pattern currently existing in the application code is to have a `useEffect` that updates the session storage when the React state value changes. An example of this is the updating of the `editMode` value used in some pages. As the value is just a toggle (albeit today only setting it to the `true` value) the code can be simplified by updating the session storage in the same React state callback function that updates the state (in parallel with it).

The existing pattern of using `useEffect` has been observed to trigger a session storage update one too many, probably related to the session storage key name being dynamic (using the location pathname). This bug was avoided by updating the session storage in parallel with the React state.

joargr:
Resolves #420 

Values persisted to session storage are updated in parallel with updates to React state, instead of being updated afterwards with `useEffect`. This is done for two pages that toggles a value (`editMode`). A consequence of this change is that the value is stored in the session storage only when the user activates edit mode. Previously a value was stored to session storage immediately when the page was visited, and even when the user never activated edit mode.

Other values are still stored to session storage with `useEffect`, as these values are updated in React state by other means (not toggling by user interaction).

## Checklist

### DIFF
--- a/frontend/src/components/ProjectRecoveryNotification.tsx
+++ b/frontend/src/components/ProjectRecoveryNotification.tsx
@@ -6,6 +6,7 @@ import { toast } from "react-toastify";
 
 import { projectGetCache, projectPostCacheRestore } from "#client";
 import {
+  projectGetChangelogQueryKey,
   projectGetLockStatusQueryKey,
   projectGetProjectQueryKey,
 } from "#client/@tanstack/react-query.gen";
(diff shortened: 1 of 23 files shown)


GibranAlfa (approved):
less useEffect the better it is!

PR #426, opened 2026-06-05
MAINT: Remove unused component, and move CSS fix to global styles

### LINKED ISSUE #425: Some fixes to the lock status banner
Some fixes and cleanup on the lock status banner should be done:

- There is a styled component named `LockStatusBanner`. This is not in use, and can be removed. Note that the component that shows the lock status banner is also named `LockStatusBanner`
- The lock status banner is using the styled component `Banner`. This is based on the EDS component `Banner`, and there are a couple of fixes that are applied when used in the application. These fixes should be defined in the global style file, similar to other EDS fixes

joargr:
Resolves #425 

Removes the unused component and moves the EDS fix to the global styles. For the fix, the attribute selector is now more precise - classname must *start with* the string and not merely *contain* it. In addition, hyphens have been added to the class names to better match the EDS class names. The hyphen has also been added to the existing EDS fix in that file.

## Checklist


frontend/src/styles/global.ts
```diff
@@ -7,7 +7,15 @@ const GlobalStyle = createGlobalStyle`
   }
 
   /* Fixes to EDS components */
-  div[class^=NativeSelect__Container] {
+
+  /* Adjust elements and ensure clean border corners */
+  div[class^=Banner__Content-],
+  div[class^=Banner__NonMarginDivider-] {
```
GibranAlfa: i think the `Banner__Content` selector is OK, but the divider one doesn’t seem to hit anymore.

the divider is rendered like this:

```html
<hr class="Divider__StyledDivider-sc-... fPraSl Banner__NonMarginDivider-sc-... JVSmp">
```

so `div[class^=Banner__NonMarginDivider-]` misses it because it’s an `hr`, not a `div`, and the class string starts with `Divider__StyledDivider`, not `Banner__NonMarginDivider`.

That means the EDS divider background comes back, which I think is the very very subtle background showing in the banner now.

Image

Maybe keep just this selector as something like:

```suggestion
  hr[class*=Banner__NonMarginDivider-] {
```

joargr: Yes, good catch. This PR only moves the fix, so then the divider fix needs to be kept like before. The addition of the element (`div`, `hr`) should be done, so as to narrow what elements are affected.

GibranAlfa (approved):

PR #433, opened 2026-06-09
ENH: Reset changelog query

### LINKED ISSUE #415: Reset changelog query when fetch returns 404 Not found
When the query for fetching changelog entries (`projectGetChangelog`) results in a 404 Not found error, an error message is sometimes shown to the user but no more is done. If there is data from a previous fetch this old data is shown (instead of the error message), although this data may not be relevant for the current context. For instance, if the user is switching from a previous project to a new project, if the new project does not have a changelog (and hence the changelog query results in a 404 Not found error), changelog data from the previous project might show, even though this data and the query result has been marked as invalidated.

The handling of a 404 Not found error for a changelog query should be modified, probably so that such an error will lead to the query being reset (using `QueryClient.resetQueries`). The code for dealing with suspense queries and error boundaries can probably remain - the modification can probably be dealt with in the `useSuspenseQuery` options.

nourinmohd:
Resolves #415 

Reset changelog query when fetch returns 404 Not found. 

## Checklist

### DIFF
--- a/frontend/src/components/home/Changelog.tsx
+++ b/frontend/src/components/home/Changelog.tsx
@@ -19,7 +19,10 @@ import { FILE_LABELS, formatEntryDescription, getTypeLabel } from "./utils";
 function Content() {
   const { data } = useSuspenseQuery({
     ...projectGetChangelogOptions(),
-    meta: { preventDefaultErrorHandling: [404] },
+    meta: {
+      preventDefaultErrorHandling: [404],
+      resetQueryOnErrorStatuses: [404],
+    },
     retry: (failureCount, queryError) =>
       !(isAxiosError(queryError) && queryError.response?.status === 404) &&
       failureCount < 3,
--- a/frontend/src/components/home/Changelog.tsx
+++ b/frontend/src/components/home/Changelog.tsx
@@ -21,7 +21,7 @@ function Content() {
     ...projectGetChangelogOptions(),
     meta: {
       preventDefaultErrorHandling: [404],
-      resetQueryOnErrorStatuses: [404],
+      resetQueryOnError: [404],
     },
     retry: (failureCount, queryError) =>
       !(isAxiosError(queryError) && queryError.response?.status === 404) &&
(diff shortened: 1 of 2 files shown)


nourinmohd:
I decided to give Gibran's suggestion using the meta approach and this is what I changed : 

1. Added reusable query meta support in` main.tsx` so queries can declare status codes that should reset stale cached data:
-> New meta field: `resetQueryOnErrorStatuses`
-> n QueryCache `onError` if the response status matches and the query currently has data, the query is reset.

2. Updated the changelog suspense query in `Changelog.tsx`. 
->  Added `resetQueryOnErrorStatuses`.

GibranAlfa (approved):
nice!

GibranAlfa:
maybe we can add a new constant for 404 in api.ts:
```ts
export const HTTP_STATUS_NOT_FOUND = 404;
```
so we avoid hardcoding 404 everywhere

joargr:
> maybe we can add a new constant for 404 in api.ts:
> 
> ```ts
> export const HTTP_STATUS_NOT_FOUND = 404;
> ```
> 
> so we avoid hardcoding 404 everywhere

Agree, and this should be done as a separate issue, as implementing it would need to update all the files where the `404` code is hardcoded. The name of the constants should also include the status code number, so for `404` the constant name should be `HTTP_STATUS_404_NOT_FOUND`.

frontend/src/main.tsx
```diff
@@ -75,6 +75,7 @@ declare module "@tanstack/react-router" {
 interface QueryAndMutationMeta extends Record<string, unknown> {
   errorPrefix?: string;
   preventDefaultErrorHandling?: Array<number>;
+  resetQueryOnErrorStatuses?: Array<number>;
```
joargr: As the values of this and the existing `preventDefaultErrorHandling` key both are arrays of HTTP status codes, the naming should be similar. The existing key does not include the term "statuses", so I suggest not including that term in the new key. The key name would thus be `resetQueryOnError`.

frontend/src/components/home/Changelog.tsx
```diff
@@ -19,7 +19,10 @@ import { FILE_LABELS, formatEntryDescription, getTypeLabel } from "./utils";
 function Content() {
   const { data } = useSuspenseQuery({
     ...projectGetChangelogOptions(),
-    meta: { preventDefaultErrorHandling: [404] },
+    meta: {
+      preventDefaultErrorHandling: [404],
+      resetQueryOnErrorStatuses: [404],
```
GibranAlfa: left this one?

joargr (approved):
Very nice!

PR #436, opened 2026-06-10
BUG: Adapt SSO token handling to MSAL 5

### LINKED ISSUE #435: MSAL 5 breaks SMDA stale-token retry
After the MSAL 5 upgrade by this PR https://github.com/equinor/fmu-settings-gui/pull/403, the SMDA stale token retry flow is broken.

Root cause: MSAL PR [#8009](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/8009/changes#diff-d8bde128bad64cb357b230e9a738968be4d62b4e3d7c78e889f3e38d4689f73f) changed the `LOGIN_SUCCESS` payload.

Before, `LOGIN_SUCCESS` emitted an `AuthenticationResult`. Now it emits `result.account`.

Our code still does:

https://github.com/equinor/fmu-settings-gui/blob/4e602eb271200ca35746e77ddd7609f32f851ec9/frontend/src/main.tsx#L295-L299

With MSAL 5, `payload.account` is undefined, so no active account is set. Later, the retry path calls `acquireTokenSilent({ scopes })` without an account, which can fail.

Fix:

- Treat `LOGIN_SUCCESS` payload as `AccountInfo`.
- Treat `ACQUIRE_TOKEN_SUCCESS` payload as `AuthenticationResult`.
- For stale token retry, call `acquireTokenSilent` with the active account, falling back to the single cached account

GibranAlfa:
Resolves #435

Adapted SSO token handling to MSAL 5 event payloads. `LOGIN_SUCCESS` now sets the active account from the `AccountInfo` payload, while `ACQUIRE_TOKEN_SUCCESS` continues to handle the `AuthenticationResult` access
token. Stale SMDA token refresh now calls `acquireTokenSilent` with the active account, falling back to the single cached account.

Tested this by manually patching the `smda_api` token with a broken token and it retried after 401.
2026-06-10_2-07-34 PM
Easiet fix is to just revert back to MSAL 4, but I figured we can just adapt to the new MSAL 5.

## Checklist

### DIFF
--- a/frontend/src/main.tsx
+++ b/frontend/src/main.tsx
@@ -1,4 +1,5 @@
 import {
+  type AccountInfo,
   type AuthenticationResult,
   type EventMessage,
   EventType,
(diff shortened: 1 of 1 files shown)


joargr (approved):
I haven't tried to simulate the situation where the SMDA token has gone stale, but it seems that this fix is ok.

PR #437, opened 2026-06-10
ENH: Improve horizon dropdown lists

### LINKED ISSUE #414: Improve the horizon dropdown lists
1. Currently the GUI implementation concatenate each mapped top zone's top and base horizons without deduplicating UUIDs in https://github.com/equinor/fmu-settings-gui/blob/b694c19170359594623e1d41ea8785387d2b7230/frontend/src/components/project/stratigraphy/Overview.tsx#L497-L508

Using Drogon data, we will see this error because a base horizon name and uuid of one zone is the same as the top horizon name and uuid of zone below it. We need to deduplicate the `horizonOptions` to not see this error. Can't find any real field to replicate, but I think this is worth fixing.

Image

2. It might be better to populate the horizon lists only from the associated zone (zone above and below it). Not sure if there's any use case that we want to map horizon which is **not** associated with the zone above and below. Wdyt @tnatt @krhart ?

GibranAlfa:
Resolves #414

- Deduped horizon options by UUID and show the full list of horizon options 
- Horizons from associated zones are listed on top. When a zone directly above or below a horizon is mapped, its horizons are showed at the top of that horizon's options and annotated with the source zone name for easier selection. This added label is similar to what we have in masterdata editing for stratigraphic column and coordinate system.

Example:

When mapping TopTherys horizon, using JS strat column as mapping example, we got Balder Fm. Base and Sele Fm. Top as suggestions on top of the list.
2026-06-10_2-47-13 PM

In Drogon, when mapping TopVolantis horizon, we got VOLANTIS GP. Top as suggested horizon, because both Valysar Fm. and VOLANTIS GP. zone have VOLANTIS GP. Top as their top horizon name.
2026-06-10_2-48-37 PM

## Checklist

### DIFF
--- a/frontend/src/components/form/field.tsx
+++ b/frontend/src/components/form/field.tsx
@@ -35,7 +35,6 @@ export interface CommonTextFieldProps
 export interface OptionProps {
   value: string;
   label: string;
-  displayLabel?: string;
 }
 
 const helperTextLoadingOptions = "Loading options...";
@@ -215,7 +214,7 @@ export function Select({
       >
         {options.map((option) => (
           <option key={option.value} value={option.value}>
-            {option.displayLabel ?? option.label}
+            {option.label}
           </option>
         ))}
       </NativeSelect>
--- a/frontend/src/components/form/field.tsx
+++ b/frontend/src/components/form/field.tsx
@@ -35,6 +35,7 @@ export interface CommonTextFieldProps
 export interface OptionProps {
   value: string;
   label: string;
+  displayLabel?: string;
 }
 
 const helperTextLoadingOptions = "Loading options...";
@@ -214,7 +215,7 @@ export function Select({
       >
         {options.map((option) => (
           <option key={option.value} value={option.value}>
-            {option.label}
+            {option.displayLabel ?? option.label}
           </option>
         ))}
       </NativeSelect>
--- a/frontend/src/components/project/stratigraphy/Overview.tsx
+++ b/frontend/src/components/project/stratigraphy/Overview.tsx
@@ -397,12 +397,6 @@ function Elements({
   const [activeElementMapping, setActiveElementMapping] = useState<
     ElementMapping | undefined
   >();
-  const [smdaNameOptions, setSmdaNameOptions] = useState<
-    Record<ElementType, OptionProps[]>
-  >({
-    horizon: [...createSpecialOptions("horizon", false)],
-    zone: [...createSpecialOptions("zone", false)],
-  } as Record<ElementType, OptionProps[]>);
   const [editDialogOpen, setEditDialogOpen] = useState(false);
   const queryClient = useQueryClient();
   const frameworkData = useFrameworkData();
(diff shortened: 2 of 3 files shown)


frontend/src/components/form/field.tsx
```diff
@@ -214,7 +215,7 @@ export function Select({
       >
         {options.map((option) => (
           <option key={option.value} value={option.value}>
-            {option.label}
+            {option.displayLabel ?? option.label}
```
joargr: I see from the code that the `label` data is used to get the horizon name, and that this is used when storing the data of the selected horizon to the API. As there is a need for an adjusted label for some of the options, this `displayLabel` property has been introduced, so as to not influence the horizon name stored to the API.

I think it's unfortunate to have a new property for this, as the existing `label` is the one where the text shown to the user is specified. It is confusing to now have both a `label` and a `displayLabel` property, as the property names both suggest that it is for showing the label, and it is not clear which one should be used and when to use which. When a form is saved, the form handling should get only the `value`(the id) from the select widget, and if more data is needed then this should be looked up in the code, not gotten from the options list. See also the comment on the code when creating the horizon options list.

frontend/src/components/project/stratigraphy/functions.ts
```diff
@@ -118,6 +119,75 @@ export function createStratUnitOptions(stratUnits: StratigraphicUnit[]) {
...
+        return;
+      }
+      seenHorizonUuids.add(horizonUuid);
+
+      const zoneNames = zoneNamesByHorizon.get(horizonUuid);
+      if (zoneNames) {
+        adjacentHorizonOptions.push({
+          value: horizonUuid,
+          label: horizonName,
+          displayLabel: `${horizonName} [${[...zoneNames].join(", ")}]`,
```
joargr: As mentioned for the `Select` component, there shouldn't be a new property for a `displayLabel`. The text to show to the user should be put in the existing `label` property, including the suffix for those horizons that have that.

As there is a need to have the horizon name for the storing of the mappings to the API, I suggest the use of a lookup. The data in this function can return this lookup in addition to the options list. The lookup can be a simple `Record<string, string>` of horizon uuid as the key and the name as the value. This lookup could then be used in the `mutationCallback`, when calling the `updatedElementMappings` function. This function would the receive the horizon name instead of the `editSmdaNameOptions` list, and the function would not need any logic for finding the horizon name from an options list. If the element being saved is a zone, the horizon name is not needed and the parameter could be the empty string.

As part of the implementation of automatic mapping in the GUI, I am doing some refactoring of the various functions and components used by the stratigraphy mapping page. There will probably be some refactoring of the functions that process the saved value, but for this PR it would be good to simplify the `updatedElementMappings` function by providing it with the horizon name.

frontend/src/components/project/stratigraphy/Overview.tsx
```diff
@@ -492,39 +493,35 @@ function Elements({
...
-          ]);
-
-      setSmdaNameOptions((options) => ({
-        ...options,
-        horizon: [
-          ...createSpecialOptions("horizon", horizonOptions.length > 0),
-          ...horizonOptions,
-        ],
-      }));
+  const editSmdaNameOptions = useMemo(() => {
```
joargr: As suggested in other comments, I don't think the horizon name should be taken from the options list when storing the updated mappings. Then there shouldn't be a need for a separate `editSmdaNameOptions` list, and the list of options for the horizon editing could be maintained in the regular `smdaNameOptions` list, as is the case prior to this PR.

frontend/src/components/project/stratigraphy/functions.ts
GibranAlfa: Good suggestion, I implemented it

joargr (approved):
This is good!

PR #443, opened 2026-06-16
BUG: Respond to closing RMS

### LINKED ISSUE #200: Close RMS when changing project
If RMS is open when changing to a new fmu project, the gui should send a request to close the opened RMS project. This will free up the rmsapi licence and prevent access to an incorrect RMS project.

Currently we ensure that users are fetching data from the correct RMS project by simply removing the `isRmsProjectOpen` session constant, which will enforce users to press the `Open project` button before they can do any edits.

### LINKED ISSUE #341: Improve the user experience when RMS project is closed when expired
There's a mismatch between the description in the page and the RMS project state in the backend.

Image

This is most likely due to `isRmsProjectOpen` is not set to `False` when RMS project is closed because it's expired. After this [PR](https://github.com/equinor/fmu-settings-api/pull/325) is merged, we can think more about improving UX when RMS session is expired. However, this issue might not be as important if we change the API to just fetch data from RMS and close it afterward. Then there's no need to deal with expired RMS session.

tnatt:
Resolves #341
Resolves #200 

PR to start using the `sessionResponse` to read the rms expiry time. This removes the need for setting rms open status in the session storage, and enables the GUI to respond to an expired rms session.

When RMS is expired 

- a dialog similar to the LockExpiryNotification is shown
- the buttons are changed from `Load/Close RMS project` to `Access RMS project` again
- opening the stratigraphy edit dialog is disabled.

## Checklist

### DIFF
--- a/frontend/src/components/RmsExpireNotification.tsx
+++ b/frontend/src/components/RmsExpireNotification.tsx
@@ -1,0 +1,190 @@
+import { Dialog } from "@equinor/eds-core-react";
+import { useMutation, useQueryClient } from "@tanstack/react-query";
+import { useEffect, useRef, useState } from "react";
+
+import {
+  rmsDeleteRmsProjectMutation,
+  rmsGetZonesOptions,
+  sessionGetSessionOptions,
+  sessionGetSessionQueryKey,
+} from "#client/@tanstack/react-query.gen";
+import { GeneralButton } from "#components/form/button";
+import { sessionRmsExpireNotificationThreshold } from "#config";
+import { useProject } from "#services/project";
+import { GenericDialog, PageText } from "#styles/common";
+
+function getSecondsUntilRmsExpiry(rmsExpiresAt?: string | null) {
+  return rmsExpiresAt
+    ? Math.max(0, Math.ceil((Date.parse(rmsExpiresAt) - Date.now()) / 1000))
+    : Number.POSITIVE_INFINITY;
+}
+
+export function RmsExpireNotification() {
+  const project = useProject();
+
+  const [isDialogOpen, setIsDialogOpen] = useState(false);
+  const isCheckingThreshold = useRef(false);
+  const [timeUntilExpire, setTimeUntilExpire] = useState<number>(
+    Number.POSITIVE_INFINITY,
+  );
+  const rmsExpiresAt = project.rmsExpiresAt;
+  const isExpired = !rmsExpiresAt;
+
+  const queryClient = useQueryClient();
+
+  const rmsRefreshMutation = useMutation({
+    mutationFn: async () => {
+      // Invalidate an rms query to trigger rms session extension.
+      await queryClient.fetchQuery({
+        ...rmsGetZonesOptions(),
+        staleTime: 0,
+      });
+    },
+    onSuccess: () => {
+      void queryClient.invalidateQueries({
+        queryKey: sessionGetSessionQueryKey(),
+      });
+      setIsDialogOpen(false);
+    },
+    meta: {
+      errorPrefix: "Error refreshing RMS access",
+    },
+  });
+
+  const rmsCloseMutation = useMutation({
+    ...rmsDeleteRmsProjectMutation(),
+    onSuccess: () => {
+      void queryClient.invalidateQueries({
+        queryKey: sessionGetSessionQueryKey(),
+      });
+      setIsDialogOpen(false);
+    },
+    meta: {
+      errorPrefix: "Error closing the RMS project",
+    },
+  });
+
+  useEffect(() => {
+    const initialTimeLeft = getSecondsUntilRmsExpiry(rmsExpiresAt);
+    setTimeUntilExpire(initialTimeLeft);
+
+    if (!Number.isFinite(initialTimeLeft) || initialTimeLeft <= 0) {
+      return;
+    }
+
+    const interval = setInterval(() => {
+      setTimeUntilExpire((currentTimeLeft) => {
+        if (!Number.isFinite(currentTimeLeft) || currentTimeLeft <= 0) {
+          return currentTimeLeft;
+        }
+
+        return currentTimeLeft - 1;
+      });
+    }, 1000);
+
+    return () => {
+      clearInterval(interval);
+    };
+  }, [rmsExpiresAt]);
+
+  useEffect(() => {
+    if (timeUntilExpire === 0) {
+      void queryClient.invalidateQueries({
+        queryKey: sessionGetSessionQueryKey(),
+      });
+    } else if (
+      !isExpired &&
+      !isDialogOpen &&
+      timeUntilExpire <= sessionRmsExpireNotificationThreshold
+    ) {
+      if (isCheckingThreshold.current) {
+        return;
+      }
+
+      isCheckingThreshold.current = true;
+
+      void queryClient
+        .fetchQuery({
+          ...sessionGetSessionOptions(),
+          staleTime: 0,
+        })
+        .then((freshSession) => {
+          const freshTimeUntilExpire = getSecondsUntilRmsExpiry(
+            freshSession.rms_expires_at,
+          );
+
+          setTimeUntilExpire(freshTimeUntilExpire);
+
+          setIsDialogOpen(
+            freshTimeUntilExpire <= sessionRmsExpireNotificationThreshold,
+          );
+        })
+        .finally(() => {
+          isCheckingThreshold.current = false;
+        });
+    }
+  }, [queryClient, timeUntilExpire, isDialogOpen, isExpired]);
+
+  const onRmsAccessRefresh = () => {
+    rmsRefreshMutation.mutate();
+  };
+
+  const onRmsAccessRelease = () => {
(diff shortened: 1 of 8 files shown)


frontend/src/components/RmsExpireNotification.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+      toast.info(
+        "Access to the RMS project has expired. " +
+          "Re-access the project to continue reading RMS data.",
+      );
+
+      return;
+    }
+
+    if (timeUntilExpire <= sessionRmsExpireNotificationThreshold) {
```
GibranAlfa: right now it will ping get session every second once timeUntilExpire is below the threshold, which is a bit unnecessary. i think it's better not to have the threshold, and just refetch session once when it expired to confirm the expiry time is correct from the backend, and then close it if it expired. if rms is not yet expired, then update timeUntilExpire, and do nothing.

frontend/src/components/RmsExpireNotification.tsx
```diff
@@ -0,0 +1,89 @@
...
+    };
+  }, [rmsExpiresAt, timeUntilExpire]);
+
+  useEffect(() => {
+    if (timeUntilExpire === 0) {
+      void queryClient.invalidateQueries({
+        queryKey: sessionGetSessionQueryKey(),
+      });
+
+      toast.info(
```
GibranAlfa: since expiry changes the user’s access state, i think a persistent dialog like `LockExpireNotification` would be easier to notice than a toast. the toast is going away so quickly, so i won't probably notice it if i come back from 2 hours lunch :)

tnatt: Agree, a dialog is now in place instead 👍

frontend/src/components/RmsExpireNotification.tsx
tnatt: yup, my mistake 😃 should work better now!

frontend/src/components/RmsExpireNotification.tsx
```diff
@@ -0,0 +1,176 @@
...
+  useEffect(() => {
+    if (timeUntilExpire === 0) {
+      void queryClient.invalidateQueries({
+        queryKey: sessionGetSessionQueryKey(),
+      });
+    } else if (
+      !isExpired &&
+      isDialogOpen &&
+      timeUntilExpire > sessionRmsExpireNotificationThreshold
+    ) {
```
joargr: Will this case ever occur? The dialog is open only when `timeUntilExpire` is less than or equal to the threshold (as set in lines 102-106), and therefore `timeUntilExpire` cannot be larger than the threshold. It should be ok to remove this check?

frontend/src/components/RmsExpireNotification.tsx
```diff
@@ -0,0 +1,176 @@
...
+      errorPrefix: "Error closing the RMS project",
+    },
+  });
+
+  useEffect(() => {
+    setTimeUntilExpire(getSecondsUntilRmsExpiry(rmsExpiresAt));
+
+    if (!Number.isFinite(timeUntilExpire) || timeUntilExpire <= 0) {
+      return;
+    }
```
joargr: I assume that the value of `timeUntilExpire` being checked is the one calculated just above, and which is updated in state. I think the check will get the previous value of `timeUntilExpire`, so there might be a problem with the check. In any case it should be better to get the updated value of `timeUntilExpire` first, and then use the value to update the state and perform the check.

frontend/src/components/RmsExpireNotification.tsx
```diff
@@ -0,0 +1,176 @@
...
+  }, [queryClient, timeUntilExpire, isDialogOpen, isExpired]);
+
+  const onRmsAccessRefresh = () => {
+    void queryClient.invalidateQueries({
+      queryKey: rmsGetZonesQueryKey(),
+    });
+    void queryClient.invalidateQueries({
+      queryKey: sessionGetSessionQueryKey(),
+    });
+  };
```
GibranAlfa, line 130: `invalidateQueries()` refetches only active queries by default. It works on the RMS page because the zones query remains mounted, but if you navigate to another page and the RMS notification appears, clicking “Continue RMS access” does not call `GET /rms/zones`, so the RMS session still expires. We can explicitly fetch the endpoint here.

Also currently zones and session are invalidated at the same time, so `GET /session` can race ahead and return the old expiry. Suggest to fetch endpoint and refresh session in sequence.

We also need to handle the loading state and disable state for the continue button and close button, even though it very fast to refresh RMS expiry and close RMS project, still good to have.
```suggestion
  const rmsRefreshMutation = useMutation({
    mutationFn: async () => {
      await queryClient.fetchQuery({
        ...rmsGetZonesOptions(),
        staleTime: 0,
      });
  
      await queryClient.invalidateQueries({
        queryKey: sessionGetSessionQueryKey(),
      });
    },
    meta: {
      errorPrefix: "Error refreshing RMS access",
    },
  });
  
  const onRmsAccessRefresh = () => {
    rmsRefreshMutation.mutate();
  };
```

frontend/src/components/RmsExpireNotification.tsx
```diff
@@ -0,0 +1,176 @@
...
+          <>
+            <GeneralButton
+              label="Continue RMS access"
+              onClick={onRmsAccessRefresh}
+            />
+            <GeneralButton
+              label="Close RMS for access"
+              variant="outlined"
+              onClick={onRmsAccessRelease}
+            />
```
GibranAlfa, line 184:
```suggestion
            <GeneralButton
              label="Continue RMS access"
              isPending={rmsRefreshMutation.isPending}
              disabled={rmsCloseMutation.isPending}
              onClick={onRmsAccessRefresh}
            />
            
            <GeneralButton
              label="Close RMS for access"
              variant="outlined"
              isPending={rmsCloseMutation.isPending}
              disabled={rmsRefreshMutation.isPending}
              onClick={onRmsAccessRelease}
            />
```

frontend/src/components/RmsExpireNotification.tsx
tnatt: Oh that I did not know about the invalidate query, thanks 🙂  good suggestion with the sequence! I'll apply it

tnatt: btw it is probably better to move the invalidation of the session response to the `onSuccess` callback instead @GibranAlfa?
```
    onSuccess: () => {
      void queryClient.invalidateQueries({
        queryKey: sessionGetSessionQueryKey(),
      });
      setIsDialogOpen(false);
    },
```

frontend/src/components/RmsExpireNotification.tsx
tnatt: Seems I can remove it, just had to add a close of the dialog when the rms access refresh was successful first 👍 
Before that removing this part would prevent the dialog from closing when the `Continue accessing RMS ` was pressed

frontend/src/components/RmsExpireNotification.tsx
tnatt: Applied

frontend/src/components/RmsExpireNotification.tsx
tnatt: I have updated this now

frontend/src/components/RmsExpireNotification.tsx
GibranAlfa: yes i think that better as well 👍

frontend/src/components/RmsExpireNotification.tsx
```diff
@@ -0,0 +1,191 @@
...
+      });
+    }, 1000);
+
+    return () => {
+      clearInterval(interval);
+    };
+  }, [rmsExpiresAt]);
+
+  useEffect(() => {
+    console.log(timeUntilExpire, isExpired, isDialogOpen);
```
GibranAlfa: do we still need this log?

tnatt: hehe

frontend/src/components/RmsExpireNotification.tsx
```diff
@@ -0,0 +1,190 @@
...
+          <>
+            <GeneralButton
+              label="Continue RMS access"
+              isPending={rmsRefreshMutation.isPending}
+              disabled={rmsCloseMutation.isPending}
+              onClick={onRmsAccessRefresh}
+            />
+
+            <GeneralButton
+              label="Close RMS for access"
```
GibranAlfa: is it better to just say "Close RMS access"?

frontend/src/components/project/rms/Overview.tsx
GibranAlfa: maybe better to just say "Close RMS access"

GibranAlfa (approved):
looks good to me :) some small comments about the button label to consider

joargr (approved):
Looks good.

frontend/src/components/RmsExpireNotification.tsx
tnatt: Yes I think so, and changed it

PR #446, opened 2026-06-17
ENH: Create GUI session before loading project state

### LINKED ISSUE #440: FMU settings sometimes opens without a project
Occasionally (not often)  FMU settings opens without a project selected, even though it was opened from a project root with a valid `.fmu`.
Hence the first thing users see is the `Select project` button, which is not a great UX. 

Can there be a race problem here, i.e. that the endpoint that returns the fmu project has not succeeded before the application is opened? 

**Note**
It has happened two times directly after `fmu init` was run on the terminal, and the last time I can't recall, but I don't think the initialization has nothing to do with this problem and guess that was just coincidental.

GibranAlfa:
Resolves #440 

- Added a simple session startup gate: once a token exists and the interceptor is installed, the GUI calls `POST /session/`.
This is how the flow of `fmu settings` command before this PR but with new change in CLI that it waits to open browser after API health is OK.
2026-06-17_10-11-53 AM
This is the flow after changes in this PR.
2026-06-17_10-07-47 AM

- Added `sessionReady` to router context and gated the root shell, so Header, Sidebar, notifications, and route content do not mount until session creation succeeds. Before session is ready, the GUI shows `Loading...`.
If session creation fails, the GUI shows a clear error.

- Kept 401 recovery as fallback. If a later API request gets 401, the existing interceptor can still request a new session. After successful session creation, user and project queries are invalidated so they refetch with the new cookie.

- Hardened `useProject()`.

  - 200 still returns project data
  - 404 still becomes “No project selected”
  - 401 still throw, then auth flow handles it, unchanged
  - 422 still supports the recovery flow, but now shows "Project configuration is invalid" text in GUI instead of “No project selected”
  - 500, network/no response, unexpected errors now throw instead of being cached as “No project selected”

- Removed `sessionCreatedAt`. This is used to retry `POST /session/` after 10s. In this case, retrying a failed `POST /session/` would not help based on current API implementation. The failure cases are invalid token, user .fmu permission/filesystem problem, backend bug, API not reachable/network failure. The first three are not helped by automatic retry. The fourth is the only case where retry might help, but the CLI `/health` wait already reduces that at startup. Also, the existing mutation retry helper already does not retry failed session creation.
## Checklist

### DIFF
--- a/frontend/src/components/Header.tsx
+++ b/frontend/src/components/Header.tsx
@@ -25,6 +25,7 @@ import { LockIcon } from "#components/LockStatus";
 import { useProject } from "#services/project";
 import { useTaskList } from "#services/tasks";
 import { GenericDialog, PageText } from "#styles/common";
+import { HTTP_STATUS_422_UNPROCESSABLE_CONTENT } from "#utils/api";
 import { AppMenu } from "./AppMenu";
 import {
   FmuLogo,
@@ -105,6 +106,8 @@ function ProjectInfo() {
             value={project.data.config.model?.revision}
           />
         </>
+      ) : project.errorStatus === HTTP_STATUS_422_UNPROCESSABLE_CONTENT ? (
+        "Project configuration is invalid"
       ) : (
         "No project selected"
       )}
--- a/frontend/src/main.tsx
+++ b/frontend/src/main.tsx
@@ -31,9 +31,11 @@ import ReactDOM from "react-dom/client";
 
 import type { Options, SessionPostSessionData, SessionResponse } from "#client";
 import {
+  projectGetProjectQueryKey,
   sessionPatchAccessTokenMutation,
   sessionPostSessionMutation,
   smdaGetHealthQueryKey,
+  userGetUserQueryKey,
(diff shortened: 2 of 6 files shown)


joargr:
> * Removed `sessionCreatedAt`. This is used to retry `POST /session/` after 10s. In this case, retrying a failed `POST /session/` would not help based on current API implementation. The failure cases are invalid token, user .fmu permission/filesystem problem, backend bug, API not reachable/network failure. The first three are not helped by automatic retry. The fourth is the only case where retry might help, but the CLI `/health` wait already reduces that at startup. Also, the existing mutation retry helper already does not retry failed session creation.

Actually, `sessionCreatedAt` was used to prevent multiple session creations happening in parallel, ensuring that there would be at least 10 s between session creation. I think the likelihood of such parallel session creation happening was low, but it was good to have that guard in place. The new `isCreatingSession` flag does the same job, so it shouldn't be a problem to remove the old `sessionCreatedAt`.

joargr (approved):
This is good, a nice improvement to the startup of the GUI.

PR #462, opened 2026-06-22
DEV: Update TypeScript language target to ES2024

### LINKED ISSUE #452: Update TypeScript language target
The TypeScript configuration sets the ECMAScript (JavaScript) target as ES2020. This should be updated to the newest possible version, which for TypeScript 5.8 is ES2024. This allows for newer language features.

In the application, the configuration is set in `tsconfig.app.json` and `tsconfig.node.json` (both referenced from `tsconfig.json`). The development environment uses Node.js, which might be at version 22, and this version seems to support ES2022 and ES2023. These versions are the ones currently specified in the configuration, and there is no need to update the language target for that environment as there isn't much code running in Node.js.

In addition, the ESLint configuration, in `eslint.config.js`, should be updated to match the TypeScript app configuration.

joargr:
Resolves #452 

- TypeScript app configuration: Updated from ES2020 to ES2024. Spelling of `target` is lowercase, and uppercase for `lib`
- TypeScript Node.js configuration: Not updated, however spelling is made lowercase
- ESLint configuration: Updated from ES2020 to ES2024, value is specified as the year. Use of `project` and `projectService` is mutually exclusive, so the best approach of using `projectService` is done

With updated language versions, there has been errors and warnings, which are also fixed in this PR. Some notes:

- `services/project.ts`: Using `queryClient` in the query function is ok, but this triggers an error. The intended solution for the error (adding `queryClient` to the query key) should not be done, so the error is set to be ignored
- `services/smda.ts`: Removing the assertion of `HealthCheck` removes some information. However, in general the errors and warnings are heeded, without making exceptions, so this change is done

## Checklist

### DIFF
--- a/frontend/eslint.config.js
+++ b/frontend/eslint.config.js
@@ -21,10 +21,9 @@ export default tseslint.config(
     ],
     files: ["**/*.{ts,tsx}"],
     languageOptions: {
-      ecmaVersion: 2020,
+      ecmaVersion: 2024,
       globals: globals.browser,
       parserOptions: {
-        project: ["./tsconfig.node.json", "./.tsconfig.app.json"],
         projectService: true,
         tsconfigRootDir: import.meta.dirname,
       },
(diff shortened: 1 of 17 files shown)


GibranAlfa (approved):
looks cleaner now that we can remove many assertions 🧹

PR #464, opened 2026-06-24
DEV: Update typescript to V6

### LINKED ISSUE #461: Update TypeScript version to 6
The application currently uses TypeScript version 5.8 (released February 2025). This should be updated to the newest version of TypeScript 6 (released March 2026). This version is a transition release to the upcoming version 7, so it can be useful to update to version 6 first.

When updating, take note of the change regarding the configuration of [`lib` and the value `DOM.Iterable`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#the-dom-lib-now-contains-domiterable-and-domasynciterable). As this value is now included as part of the `DOM` value, and the `DOM` value as well as the `ES2024` value is automatically set, there is no need to specify the `lib` value, so this setting can be removed from `tsconfig.app.json`.

nourinmohd:
Resolves #461 

This PR updates the frontend TypeScript toolchain to TypeScript 6 and applies the related package/config changes needed for the transition release.

It also updates React lint-related packages and adjusts affected React state code so the newer `react-x `rules remain enabled and pass cleanly.

**Dependency Updates:** 
Updated TypeScript and related tooling in `package.json`

- typescript
- ~~@hey-api/openapi-ts~~
- eslint-plugin-react-dom
- eslint-plugin-react-x

**TypeScript Config**
Removed the explicit `lib` setting from `tsconfig.app.json`

TypeScript 6 now includes` DOM.Iterable` as part of `DOM`, and the default` DOM/ES2024 libs` cover the previous explicit config. This addresses the TypeScript 6 release-note change around `DOM.Iterable`.

**React-X Lint Changes**
The updated `eslint-plugin-react-x` usage and state updates inside effects. This PR keeps those rules enabled and updates affected code paths instead of disabling the rules globally.

**Key Points :** 
- Setter naming was aligned with `react-x/use-state`. 
- Some state reset logic was changed to derived state instead of effect-driven state
- Some effect-triggered state transitions were deferred with `Promise.resolve().then(...) `to avoid synchronous state updates directly inside effect setup
- Stratigraphy mappings were changed from effect-initialized state to keyed state

`pnpm install` may still show a TypeScript peer warning from transitive `tsconfck@3.1.6` through `vite-tsconfig-paths ` `vite-tsconfig-paths` is already latest, and `tsconfck` does not currently publish a TypeScript 6-compatible peer range.

## Checklist


frontend/package.json
```diff
@@ -33,7 +33,7 @@
   },
   "devDependencies": {
     "@eslint/js": "^9.39.2",
-    "@hey-api/openapi-ts": "^0.66.7",
+    "@hey-api/openapi-ts": "^0.98.2",
```
joargr: This package is updated, however, the changes for such an update are comprehensive, so the update should be done as a separate task. I have created an [issue](https://github.com/equinor/fmu-settings-gui/issues/471) for this.

frontend/package.json
```diff
@@ -43,12 +43,12 @@
     "@types/react-dom": "^19.2.3",
     "@vitejs/plugin-react": "^4.4.1",
     "eslint": "^9.39.4",
-    "eslint-plugin-react-dom": "^2.7.2",
+    "eslint-plugin-react-dom": "^5.9.0",
     "eslint-plugin-react-hooks": "^5.2.0",
     "eslint-plugin-react-refresh": "^0.4.20",
-    "eslint-plugin-react-x": "^1.49.0",
+    "eslint-plugin-react-x": "^5.9.0",
```
joargr, line 49: `eslint-plugin-react-dom` and `eslint-plugin-react-x` are included in the unified plugin [`@eslint-react/eslint-plugin`](https://www.npmjs.com/package/@eslint-react/eslint-plugin). There are some additional plugins that could be relevant, so I suggest that this unified plugin is specified as a dependence instead, so that additional plugins can be used. This would probably be a bigger change, so it can be created and implemented as a separate issue. The update of the two plugins in this PR and the associated code changes can be left as-is.

frontend/tsconfig.app.json
joargr, line 1: With the update to a new major version, I think it's natural to visit the TypeScript configuration as well. TypeScript 5.9 made changes to the [initial configuration](https://devblogs.microsoft.com/typescript/announcing-typescript-5-9/#minimal-and-updated-tsc---init) (as created by the command `tsc --init`), creating a `tsconfig.json` with modern configuration. Along with changes in TypeScript 6 and some guides, this can be used as a reference to update the TypeScript configuration in FMU Settings GUI.

Suggest doing the following changes in `compilerOptions` in `tsconfig.app.json`:

- [`rootDir`](https://www.typescriptlang.org/tsconfig/#rootDir): In TypeScript 6, this now [defaults to `.`](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#rootdir-now-defaults-to-.), which is the directory where the `tsconfig.json` file resides. Source files are located in `./src`, so it appears that this option needs to be set, however development seems to work like normal without this setting. It should probably only be set if there are errors when developing or building the application
- [`module`](https://www.typescriptlang.org/tsconfig/#module): Change from `ESNext` to `preserve`. Both values are recommended when using a bundler (ie. Vite), but `preserve` seems to be the better value
- [`target`](https://www.typescriptlang.org/tsconfig/#target): Change from `es2024` to `esnext`. It should be ok to go with configuration that says the latest version, as the application will run on evergreen browsers
- [`types`](https://www.typescriptlang.org/tsconfig/#types): In TypeScript 6, this now [defaults to `[]`](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#types-now-defaults-to-[]). I don't think the application needs to specify any types like this, so the empty array is probably ok. And as it is the default value, there is no need to explicitly set this option
- `sourceMap`, `declaration` and `declarationMap`: As `noEmit` is `true` these options are irrelevant
- Bundler mode:
   - [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax): Set to `true`
   - [`isolatedModules`](https://www.typescriptlang.org/tsconfig/#isolatedModules): This option is by default `true` when `verbatimModuleSyntax` is `true`, so the option can and should be removed from the configuration, as the default value is the needed value
   - [`skipLibCheck`](https://www.typescriptlang.org/tsconfig/#skipLibCheck): Set to `true`
- Linting options:
  - [`strict`](https://www.typescriptlang.org/tsconfig/#strict): The default value is `true`, so setting this explicitly is not needed 
  - [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess): Set to `true`
  - [`exactOptionalPropertyTypes`](https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes): Set to `true`
  - [`noUncheckedSideEffectImports`](https://www.typescriptlang.org/tsconfig/#noUncheckedSideEffectImports): The default value is `true`, so setting this explicitly is not needed

The `tsconfig.node.json` is the configuration when developing a Nodejs application, so that file is not relevant for FMU Settings GUI, and it can be ignored in this PR.

As the TypeScript configuration is set for the latest ECMAScript/JavaScript version, the ESLint configuration should be updated accordingly. In `eslint.config.js`, `languageOptions.ecmaVersion` is currently set to `2024`. The default value of this [option](https://eslint.org/docs/v9.x/use/configure/language-options#specifying-javascript-options) is `latest`, which is the needed value, so the option can be removed from the configuration.

nourinmohd:
Implemented the PR review follow-up.

- Changed `package.json` and `pnpm-lock.yaml` to revert `@hey-api/openapi-ts` back to `^0.66.7`, leaving the larger generator upgrade for the separate issue. 
- Updated `tsconfig.app.json `with the modern TS config changes `target: esnext`, `module: preserve`, `verbatimModuleSyntax`, `skipLibCheck`, `noUncheckedIndexedAccess`, `exactOptionalPropertyTypes`and removed `ecmaVersion` from `eslint.config.js`.

Because `exactOptionalPropertyTypes` and `noUncheckedIndexedAccess` expose real type issues, I also updated the affected frontend code to omit optional generated/API props when absent, widen internal optional prop types where the app intentionally passes undefined, and guard indexed lookups in `form`, `history`, `masterdata`, `mapping`, `RMS`, `stratigraphy`, and `route components`.

frontend/tsconfig.app.json
joargr: As the default value of `strict` is `true`, suggest removing this from the configuration. This is similar to what is done with other settings where there are default values.

frontend/src/components/project/masterdata/Edit.tsx
GibranAlfa: surfacing the warning from ci-lint

/home/runner/work/fmu-settings-gui/fmu-settings-gui/frontend/src/components/RmsExpireNotification.tsx
Warning:   69:5  warning  Do not call the 'set' function 'setTimeUntilExpire' of 'useState' synchronously in an effect. This can lead to unnecessary re-renders and performance issues  react-x/set-state-in-effect

/home/runner/work/fmu-settings-gui/fmu-settings-gui/frontend/src/components/project/masterdata/Edit.tsx
Warning:   445:5  warning  Do not call the 'set' function 'setSmdaFields' of 'useState' synchronously in an effect. This can lead to unnecessary re-renders and performance issues  react-x/set-state-in-effect

joargr: Yes, these two warnings should be followed up, with code fixes.

frontend/src/components/RmsExpireNotification.tsx
```diff
@@ -65,10 +77,7 @@ export function RmsExpireNotification() {
   });
 
   useEffect(() => {
-    const initialTimeLeft = getSecondsUntilRmsExpiry(rmsExpiresAt);
-    setTimeUntilExpire(initialTimeLeft);
-
-    if (!Number.isFinite(initialTimeLeft) || initialTimeLeft <= 0) {
+    if (getSecondsUntilRmsExpiry(rmsExpiresAt) <= 0) {
```
GibranAlfa: The state initialization fixes the lint warning, but this effect dropped the previous `Number.isFinite` check. When there is no active RMS session, `getSecondsUntilRmsExpiry()` returns `Infinity`, so the interval runs every second without changing the state. We should keep the earlier guard here.
```suggestion
    const initialTimeLeft = getSecondsUntilRmsExpiry(rmsExpiresAt);

    if (!Number.isFinite(initialTimeLeft) || initialTimeLeft <= 0) {
```

GibranAlfa (approved):
all comments are addressed, so good to go 👍 , just small regression cleanup before merging

PR #469, opened 2026-06-24
ENH: Use field UUID for masterdata lookup

### LINKED ISSUE #409: Use field uuid instead of identifier for masterdata search
GUI should show field identifier and country after doing a field search. The idea of showing country is to make it easier for the user to select a unique field to search on if a same field identifier exists in many countries. However, we will not use this country information to search for masterdata. We will use the field identifier and field uuid to search for masterdata, because uuid is unique. We already have this field uuid information from the field search result from the API, but now the API also support searching masterdata with field uuid from this PR (https://github.com/equinor/fmu-settings-api/pull/345).

GibranAlfa:
Resolves #409

- Use the selected field UUID when retrieving and grouping masterdata so fields with the same identifier remain distinct
- Masterdata lookups now retry only connection problem and `503` failures
- Previously, if fetching masterdata for a searched field failed, that failed request stayed in the editor and blocked later searches and add/remove actions. Now failed searches for fields not already saved in the project are cleared automatically, so the user can continue working without closing and reopening the editor. For example, users can select both RAIA fields, but only RAIA Brazil is added because its lookup succeeds and RAIA Angola is skipped because it has no coordinate system. Saved project fields are kept to avoid editing with incomplete masterdata.
2026-06-24_3-07-55 PM

## Checklist

- [x] Tests added (if not, comment why) — not relevant

### DIFF
--- a/frontend/src/components/project/masterdata/Edit.tsx
+++ b/frontend/src/components/project/masterdata/Edit.tsx
@@ -9,6 +9,7 @@ import {
 import { arrow_back, arrow_forward } from "@equinor/eds-icons";
 import { createFormHook } from "@tanstack/react-form";
 import { useMutation, useQueries, useQueryClient } from "@tanstack/react-query";
+import { isAxiosError } from "axios";
 import {
   type Dispatch,
   type SetStateAction,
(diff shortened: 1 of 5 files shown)


tnatt (approved):
Based on local testing this seems to work as expected 👍  Nice that errors from the API are no longer hidden!

PR #477, opened 2026-07-07
ENH: Restucture menu

### LINKED ISSUE #467: Restructure menu bar to have subpages for stratigraphy and wellbores
The menu bar to the left in FMU Settings should be restructured so that `stratigraphy` and `wellbores` are subpages to both `RMS` and `Mappings` (today called Stratigraphy).

The menu should look something like this

``` yml
Project
    RMS
         Overview 
         Stratigraphy
         Wellbores
    Mappings
         Overview 
         Stratigraphy
         Wellbores
User
```
In addition: 
* The RMS.Stratigraphy page should contain the stratigraphy block on the RMS page
* A new page should be created for RMS.wellbores (will be filled with content in a different issue)
* A new page should be created for RMS.overview (will be filled with content in a different issue)
* The Mappings.Stratigraphy page should point to what today is the Project.Stratigraphy page
* A new page should be created for Mappings.wellbores (will be filled with content in a different issue)
* A new page should be created for Mappings.overview (will be filled with content in a different issue)

nourinmohd:
Resolves #467 

Restructured the `Project` sidebar navigation so `RMS` and `Mappings` are grouped under `Project`, each with `Overview`, `Stratigraphy`,  and `Wellbores` subpages.

- Added nested routes for:
      `/project/rms/overview`
       `/project/rms/stratigraphy`
       `/project/rms/wellbores`
       `/project/mappings/overview`
        `/project/mappings/stratigraphy`
       `/project/mappings/wellbores`

- Moved the existing RMS stratigraphy block from the RMS overview page to the new RMS Stratigraphy page.
- Moved the existing Project Stratigraphy page behavior to the new Mappings Stratigraphy route.
- Added placeholder pages for RMS Overview, RMS Wellbores, Mappings Overview, and Mappings Wellbores where future content will be added.
- Kept old route entry points working by redirecting:
     -`/project/rms to /project/rms/overview`
     - `/project/stratigraphy to /project/mappings/stratigraphy`
- Updated task-list links to point to the new route structure.
- Updated/generated the TanStack Router route tree.
- Adjusted the sidebar styling so RMS and Mappings appear as nested groups inside Project rather than separate top-level sidebar sections.

## Checklist


nourinmohd:
Just for quick look without having to jump start fmu-settings app. 

- Nested version (the one in PR) 
Screenshot 2026-07-07 at 10 42 26

- Non-Nested version with the folder icon ( just to see how it looks) 
Screenshot 2026-07-07 at 10 14 19

- Blank icon and non-nested version
image

nourinmohd:
Update from second commit : 

Changed `Sidebar.style.ts`, removing the EDS accordion borders and manually padding nested links.

Changed `Sidebar.tsx` so nested `RMS/Mappings` groups render as real `EdsSideBar.Accordions` with a typed blank icon, while preserving the existing project routes and Masterdata/History items.

GibranAlfa (commented):
**Make RMS and Mappings work the same way**
Small thing: RMS is set up as a nested route with a parent, but Mappings isn't, its pages are just floating at the top level. Would be nice to make them match. If `rms.tsx` becomes a simple wrapper (container + `<Outlet/>`) and we add a matching `mappings.tsx`, both groups behave the same and `/project/mappings` gets a real landing page like RMS does. Like this in `routes/project/mappings.tsx`:
```ts
import { createFileRoute, Outlet } from "@tanstack/react-router";

import { PageContainerNotWidthConstrained } from "#styles/common";

export const Route = createFileRoute("/project/mappings")({
  component: RouteComponent,
});

function RouteComponent() {
  return (
    <PageContainerNotWidthConstrained>
      <Outlet />
    </PageContainerNotWidthConstrained>
  );
}
```
`routes/project/rms.tsx` should be edited to be similar like this.

**Simpler redirect**
For sending `/project/rms` to `/project/rms/overview`, instead of checking the pathname in the parent, we could just add a tiny `index.tsx` in `routes/project/mappings` that does `<Navigate to="…/overview" replace />`, same thing you already used in `stratigraphy.tsx.` It's cleaner and won't break if there's a trailing slash. This is similar with how this codebase is structured now. Example for `routes/project/mappings/index.tsx`
```ts
import { createFileRoute, Navigate } from "@tanstack/react-router";

export const Route = createFileRoute("/project/mappings/")({
  component: () => <Navigate to="/project/mappings/overview" replace />,
});
```
then it should be similar for `routes/project/rms/index.tsx`


**Headers repeat the section name**
I know we discussed about this, but I think it's cleaner to not repeat the "RMS" and "Mappings" in the header because the sidebar already tells you which section you're in. Could we just have each page say "Overview", "Stratigraphy", etc.?

**Redundant stratigraphy headings on the RMS Stratigraphy page**
The `Stratigraphy` component has its own heading, so now that it's on its own page you get the title twice. Since it used to sit inside RMS Overview, we can just drop the heading inside the component and keep the page one.

**Split the RMS checklist task**
"Set RMS project and stratigraphy" only links to Overview, but stratigraphy has its own page now. Might be clearer to make it two tasks: "Set RMS project" (Overview) and "Set RMS stratigraphy" (Stratigraphy), and each one should redirect you to the correct page.

**Empty pages looks broken**
The placeholder pages say "No … content available yet", which kind of looks broken. Could we switch them to a "coming soon" message so it's obviously on purpose?

**Add comments on the EDS Accordion hacks**
We should add comments on the `NestedAccordion` styles (why we remove the borders and indent) and the `blankIcon` (EDS requires you to pass an icon, and a blank one keeps the labels lined up).

frontend/src/routes/project/mappings.tsx
```diff
@@ -0,0 +1,15 @@
...
+export const Route = createFileRoute("/project/mappings")({
+  component: RouteComponent,
+});
+
+function RouteComponent() {
+  return (
+    <PageContainerNotWidthConstrained>
+      <Outlet />
+    </PageContainerNotWidthConstrained>
+  );
```
GibranAlfa: sorry I made the wrong example earlier, in the parent page, we should not wrap everything in `<PageContainerNotWidthConstrained>`, because it will make the page use the full width. We only want the child page to intentionally use it, like the stratigraphy page, so we can expand the table to use the full width of the page. so this should be removed like this
```suggestion
  return <Outlet />;
```

frontend/src/routes/project/rms.tsx
```diff
@@ -1,46 +1,15 @@
...
-  ) : (
-    <PageSectionWidthConstrained>
-      <PageText>Project not set.</PageText>
-    </PageSectionWidthConstrained>
-  );
-}
-
 function RouteComponent() {
   return (
     <PageContainerNotWidthConstrained>
```
GibranAlfa: remove this as well

frontend/src/routes/project/mappings/overview.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export const Route = createFileRoute("/project/mappings/overview")({
+  component: RouteComponent,
+});
+
+function RouteComponent() {
+  const project = useProject();
+
+  return (
+    <PageContainerNotWidthConstrained>
```
GibranAlfa: remove this because only stratigraphy page should use the full width

frontend/src/routes/project/mappings/wellbores.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export const Route = createFileRoute("/project/mappings/wellbores")({
+  component: RouteComponent,
+});
+
+function RouteComponent() {
+  const project = useProject();
+
+  return (
+    <PageContainerNotWidthConstrained>
```
GibranAlfa: remove this as well, we don't know yet if wellbore page will use the full width

frontend/src/routes/project/rms/overview.tsx
```diff
@@ -0,0 +1,46 @@
...
+  ) : (
+    <PageSectionWidthConstrained>
+      <PageText>Project not set.</PageText>
+    </PageSectionWidthConstrained>
+  );
+}
+
+function RouteComponent() {
+  return (
+    <PageContainerNotWidthConstrained>
```
GibranAlfa: remove this

frontend/src/routes/project/rms/wellbores.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export const Route = createFileRoute("/project/rms/wellbores")({
+  component: RouteComponent,
+});
+
+function RouteComponent() {
+  const project = useProject();
+
+  return (
+    <PageContainerNotWidthConstrained>
```
GibranAlfa: remove this

frontend/src/routes/project/stratigraphy.tsx
GibranAlfa, line 1: we can remove this file now, as there's nothing pointing to project/stratigraphy anymore. remember to run `pnpm dev` or `pnpm build` after deleting it so the `routeTree.gen.ts` is regenerated

GibranAlfa (approved):
very nice restructuring!

slangeveld:
@joargr should probably also review this when he is back as he had some thoughts on how he wanted this, but he can do that in hindsight when he is back

GibranAlfa:
for reference this is how it looks now:
image

PR #480, opened 2026-07-10
ENH: Add RMS wellbore page

### LINKED ISSUE #466: Wellbore mappings: Add section for well selection in RMS page
To be able to do well mappings from the mappings page in FMU Settings, we need a way to select which RMS wells that should be mapped.

We can do this in a similar way as how we do it for stratigraphy in RMS, from the RMS page in FMU Settings.

To do this, we should: 
* Add a new section for wells on the RMS page. 
   - The RMS wells should be read from RMS when opening the RMS project
* Present the RMS wells to the user
* Give the user the option to select which RMS wells they want to make available for mapping. Maybe the easiest here is to just autoselect all mappings and let the user de-select the wells that should not be mapped?
* Give the user a way to set if a well is multilateral
* Give the user a way to set which wells are planned or not

A possible feature for later, but not in this first iteration
* Include and option to read RMS wells from csv file and preselect those for the user

GibranAlfa:
Resolves #466 

- Added the RMS wellbore page
- Automatically selects all RMS wells when no wells are stored, while still requiring an explicit save.
- Added persistent filter button in Well column in the Edit dialog, to let user search for a well and then uncheck it from included well.
- Well and Planned column are filterable and sortable.
- Added virtualized, height-capped grids and styling so it's optimized for thousands of wells.
- Extracted the orphan warning into a shared component and reused it for stratigraphy and wellbore page. Also offer a button to remove them, because previously there's no way to do it.
- Added scroll handling for large edit dialogs.
- `EDSDataGrid` has some specific overrides for Firefox, so some styling adjustments are required.
- Added a new checklist

Note: multilateral well is not yet supported because current `RmsWell` model does not support that. Need discussion in the team if we need to support this.
image
image
2026-07-10_1-13-27 PM
image

## Checklist


krhart:
I think the page looks nice! It's easy to understand what to do. However, the filter could be improved a little.

- I like the orphan warning and the button to remove outdated elements 👌


- Could we have a color coding or something to make is more visible which are Planned or not (the Yes/No column), if you have a long list its not easy to separate them.
image

- When you open the Filter and filter on something, and then select one, the filtered list is reset, so you have to filter again and again for each one you want to tick off
-image

- Could there be some kind of symbol or message showing that a list is filtered? Not just the color change of the Filter icon, which I think is quite easy to miss. Or maybe make the icon red or something if it is filtered? Now the page shows the filtered list if there is a filter added but does not say that anywhere. The text at the top just says how many are Included
image

- If all wells are selected and I open the Filter to deselect just a few, none of the wells in the filtered list are selected.
image

GibranAlfa:
Thanks for testing this! I have made some changes based on your feedback:

1. Planned wells now have a highlighted row background, making them easier to distinguish in longer lists. The Planned value is still shown so the meaning does not depend on color alone.
2. I replaced the column filter with a text filter above the table. It filters by partial well name, and the filter remains active when you include or exclude wells. You no longer need to enter the filter again for each well.
3. When a filter is active, the page now shows "Showing X of Y wells", so it is clear that only part of the list is visible.
4. You were right that the checkboxes were confusing. The new text filter removes those filter checkboxes. The tip also now explicitly says to clear checkboxes in the Include column to exclude wells.
5. I also renamed the bulk actions to "Select all wells" and "Deselect all wells" to make it clear that they affect all wells, including wells hidden by the filter.

frontend/src/components/project/rms/Wellbores.tsx
```diff
@@ -0,0 +1,624 @@
...
+    setIsDialogOpen(true);
+  };
+
+  return (
+    <>
+      <PageSectionWidthConstrained>
+        <PageText>
+          The following wells are stored in the project. Planned wells are
+          excluded from wellbore mapping. All other stored wells are available
+          for mapping.
```
joargr: Is it deliberate that both terms "well" and "wellbore" are used? The page is titled "Wellbores", and the mapping will be done for "wellbores", but the text talks about "wells". If "wellbore" is the correct term then perhaps the text should be consistent in using the same term.

frontend/src/routes/project/rms/wellbores.tsx
```diff
@@ -11,17 +14,32 @@ export const Route = createFileRoute("/project/rms/wellbores")({
...
-      </PageText>
+      <PageText>Project not set.</PageText>
     </PageSectionWidthConstrained>
   );
 }
+
+function RouteComponent() {
+  return (
+    <>
+      <PageSectionWidthConstrained>
```
joargr: There is no need to use the `PageSectionWidthConstrained` component on this page. The way things are set up, when a page contains content that can expand to use the full width of the browser, then the `PageContainerNotWidthConstrained` component is used on the outer level (first level of the route component). Then, the parts of the page that will not have expanded content will be wrapped in the `PageSectionWidthConstrained` component. For now, only the RMS Stratigraphy (`routes/project/rms/stratigraphy.tsx`) and Mappings Stratigraphy (`routes/project/mappings/stratigraphy.tsx`) pages have expandable content.

The `PageSectionWidthConstrained` component can be removed from this file and the other files which are called from here. There are other pages as well which currently are using this component (after the menu restructuring), but these can be updated in a separate issue.

frontend/src/components/project/rms/Wellbores.tsx
GibranAlfa: Good point. The endpoint currently returns the parent RMS well name, so technically the selected object is a well. However, that name is also the root wellbore name and is used as the wellbore identifier in mappings. From the user’s perspective, the selected item is therefore a wellbore.

We can later extend the API to return all names from `well.all_wellbores`, which will include sidetracks. Using “wellbore” consistently in the UI is therefore correct for both the current behavior and that future support. I will update the visible text accordingly.

frontend/src/components/project/rms/Wellbores.tsx
```diff
@@ -0,0 +1,624 @@
...
+                rowClass={(row) => (row.original.planned ? "planned-row" : "")}
+                enableSorting
+                enableColumnFiltering
+              />
+            </WellsContainer>
+          </PageSectionWidthConstrained>
+        </>
+      ) : (
+        <PageSectionWidthConstrained>
+          <PageCode>No wells currently stored in the project.</PageCode>
```
joargr: Suggest the following rephrasing:

```suggestion
          <PageCode>No wells are currently stored in the project.</PageCode>
```

frontend/src/components/project/rms/Wellbores.tsx
```diff
@@ -0,0 +1,624 @@
...
+  });
+
+  const form = useAppForm({
+    defaultValues: {
+      wells: projectWells,
+    },
+    onSubmit: ({ value, formApi }) => {
+      if (!projectReadOnly) {
+        mutationCallback({
+          formValue: value.wells,
```
joargr: I think it would be better that the complete form value `value` is sent to the mutation callback. Then, in `mutationCallback`, the value used in `body` would be specified as `formValue.wells`. With this pattern, the submit handler simply sends the complete form value on to the mutation callback, and it is this function that knows that only a part of the of the form value is needed. This is the pattern used for other forms that use a mutation callback, such as on the RMS Stratigraphy page.

frontend/src/components/project/rms/Wellbores.tsx
```diff
@@ -0,0 +1,624 @@
...
+        isOpen={confirmClose.confirmCloseDialogOpen}
+        handleConfirmCloseDecision={confirmClose.handleDecision}
+      />
+
+      <EditDialog
+        open={isDialogOpen}
+        isDismissable={true}
+        onClose={confirmClose.handleCloseRequest}
+        $minWidth="36em"
+        $maxWidth="36em"
```
joargr: Since the minimum width and the maximum width is set to the same, it would be better to use `$width` instead.

krhart:
I think the filter function is a lot better now @GibranAlfa, it's more intuitive 👍

Would it be possible to have a "select all" function on a filtered list? If you have 1000 of wells and would like to select groups of them. Like in this screen shot, if I want to select all PSD wells I have to click one by one. Thats ok for 6 or 7 but annoying for 60 or 70... 
(Or add some kind of shortcut, like you have in office tools etc with holding the shift key to select multiple rows 😊)
image

krhart:
Could we swap the background color for planned wells and not planned wells? At least in my (old) head it makes more sense that the ones that are mappable (the No wells) should be highlighted

krhart:
Number of planned wells will probably be very few, compared to the mappable wells, so having just a few blue rows is probably better then having almost everything highlighted in blue. So I hereby withdraw my above comment 😊

frontend/src/components/project/rms/Wellbores.tsx
```diff
@@ -0,0 +1,624 @@
...
+  );
+
+  const orphanWellNames = projectWells
+    .filter((well) => !availableNames.has(well.name))
+    .map((well) => well.name);
+  const hasOrphans = orphanWellNames.length > 0;
+  const includedCount = projectWells.filter((well) =>
+    availableNames.has(well.name),
+  ).length;
+  const normalizedWellFilter = wellFilter.trim().toLocaleLowerCase();
```
joargr: If the conversion to lowercase should be done in a locale-aware fashion, then this should be done consistently through the application. The function `stringCompare` in `utils/string.ts`, which is used for instance when sorting the names of fields (assets), specifies the Norwegian locale (`no`). Thus, lower-casing the filter term should be done using the same locale.

If the locale is added to the lower-casing method, then it would be best to define the application locale `no` as an application constant, in `config.ts`.

frontend/src/components/project/rms/Wellbores.tsx
```diff
@@ -0,0 +1,624 @@
...
+        <WellSearch
+          placeholder="Filter wells"
+          value={wellFilter}
+          onChange={(event) => {
+            setWellFilter(event.target.value);
+          }}
+        />
+        {normalizedWellFilter && (
+          <PageText $marginBottom="0">
+            Showing <span className="emphasis">{visibleWells.length}</span> of{" "}
```
joargr: Maybe the text could be "Filter is showing x of y wells.", to make it more clear that what is shown is a sub-selection of the total amount of wellbores ("x of y RMS wells are included.").

joargr:
When a wellbore is selected marked as planned, the "planned" flag is reset when the wellbore is deselected. So if the user has selected a wellbore and marked it as planned, this information gets lost if the user accidentally deselects the wellbore, or accidentally clicks the "Deselect all wells" button. When the wellbore is selected again, the "planned" flag is not set.

Is it possible to keep such information even when a wellbore is deselected? Perhaps it should be possible to set the "planned" flag irrespectively of whether the wellbore is selected or not. If starting with all wellbores deselected, the user can mark the planned wellbores first, and then decide which wellbores should be selected.

frontend/src/components/project/rms/Wellbores.tsx
```diff
@@ -0,0 +1,624 @@
...
+            orphanWellNames.length === 1 ? "well stored" : "wells stored"
+          } in the project ${
+            orphanWellNames.length === 1 ? "is" : "are"
+          } no longer available in RMS. Remove ${
+            orphanWellNames.length === 1 ? "it" : "them"
+          } before saving.`}
+          listItems={orphanWellNames}
+          buttonLabel="Remove outdated wells"
+          onRemove={removeOrphans}
+        />
```
joargr: When there are orphans and the user selects "Remove outdated wells", this is a change that still needs to be saved so that the orphans are removed from the project as stored on disk. If the user selects the remove button, the dialog will no longer mention any orphans. If the user then cancels the editing, and confirms closing the dialog, the orphans are still present in the project. It may not be clear to the user that to actually remove the orphans from the project means having to click the save button as well.

There can be orphans when editing several types of data - wellbores, stratigraphy data and masterdata. There are now two approaches when informing the user of such orphans:
1. Showing the list of orphans and saying that these will be automatically removed from the project when saving the edit form
2. Showing the list of orphans and letting the user remove the orphan from the data that is being edited. The user then needs to click on the save button to actually remove the orphans from the project

I think it's a disadvantage of the second approach that the user needs to do two actions to remove the orphans from the project. Should the user click just the remove button, and then cancel the editing, then the orphans haven't actually been removed from the project.

A disadvantage of the first approach is that the save button may be disabled when there otherwise are no changes to the data being edited. However, this can be fixed by enabling the save button when there are orphans. Then the user will see the warning box about the orphans, and can click the save button to remove the orphans from the project.

Perhaps there should be a discussion in the team about these two approaches. I think the first approach is the best, as it informs the user about the orphans and says that the user doesn't have to do something extra to remove the orphans - this will happen automatically using the regular save button.

frontend/src/components/project/rms/Wellbores.tsx
```diff
@@ -0,0 +1,624 @@
...
+            <PageText>
+              Could not load wells from RMS. Reload the RMS project and try
+              again.
+            </PageText>
+          ) : availableWellsQuery.data.length === 0 &&
+            projectWells.length === 0 ? (
+            <PageText>
+              No wells are available in RMS. Add wells to the RMS project, then
+              reload the RMS project.
+            </PageText>
```
joargr: I think this should only check the number of items from the RMS project (`availableWellsQuery.data`). If there are no wellbores in the RMS project but there are wellbores in the project (orphans), this message should show. Also, the orphans would be handled by the orphan warning box, offering the user a way to fix the orphans even if the RMS project doesn't have wellbores.

frontend/src/components/project/rms/Wellbores.tsx
```diff
@@ -0,0 +1,624 @@
...
+
+  return (
+    <>
+      {hasOrphans && (
+        <OrphanWarningBox
+          message={`${orphanWellNames.length} ${
+            orphanWellNames.length === 1 ? "well stored" : "wells stored"
+          } in the project ${
+            orphanWellNames.length === 1 ? "is" : "are"
+          } no longer available in RMS. Remove ${
```
joargr: Perhaps a phrase like "... well stored in the project is currently not available in RMS" should be used, instead of "... no longer available". Maybe there are cases where a wellbore was never available in the RMS project. Showing this as "currently not available" is less specific than "no longer available", and is a better representation of what is seen by the application.

GibranAlfa:
Changes after review:

- Changed the visible terminology from “well” to “wellbore”.
- Removed unnecessary `PageContainerNotWidthConstrained` wrappers.
- Added locale-aware filtering using the shared Norwegian application locale.
- Changed filter text to "Filter is showing x of y wellbores".
- Added buttons to select or deselect all filtered wellbores shown by the current filter @krhart. If filter is not used, the buttons will turn into select or deselect all wellbores.
- Kept Planned values when wellbores are deselected and selected again.
- Added clearer tips for bulk and individual selection.
- Changed orphan cleanup so outdated items are removed when the user saves. This is also applied to stratigraphy.
- Enabled Save when orphan cleanup is the only change.
- Showed the empty RMS message together with the orphan warning when applicable.
- Changed the orphan text to “currently not available in RMS”.
- Applied the suggested form-value, dialog-width, and empty-state changes.
- Reverted a regression that made project checklist in homepage cannot be opened again after completing all the tasks.

frontend/src/components/project/rms/Wellbores.tsx
GibranAlfa: I agree with the first approach, and now the save button is enabled when there are orphans.

frontend/src/components/project/rms/Wellbores.tsx
GibranAlfa: Updated. The empty RMS message now depends only on `availableWellsQuery.data`. When stored orphan wellbores exist, the dialog also shows the orphan warning and enables Save, which removes the orphans from the stored project.

frontend/src/components/project/rms/Wellbores.style.ts
```diff
@@ -0,0 +1,42 @@
+import { Search } from "@equinor/eds-core-react";
+import { tokens } from "@equinor/eds-tokens";
+import styled from "styled-components";
+
+export const WellsContainer = styled.div`
```
joargr: As these components are used for displaying and editing wellbores, it would be good to also rename them (`WellboresContainer`, `WellboreFilterContainer`, `WellboreSearch`), for more naming consistency also for the code. Similar renaming of variables could be done in other files as well, where appropriate, to make it clear that the items being shown and edited are wellbores and not wells.

frontend/src/components/project/rms/Wellbores.tsx
```diff
@@ -0,0 +1,659 @@
...
+              ? "wellbore stored"
+              : "wellbores stored"
+          } in the project ${
+            orphanWellNames.length === 1 ? "is" : "are"
+          } currently not available in RMS. ${
+            orphanWellNames.length === 1 ? "It" : "They"
+          } will be removed when you save.`}
+          listItems={orphanWellNames}
+        />
+      )}
```
joargr: Suggest that orphan warning box be placed under the wellbores table, below the select/deselect buttons and above the tips. Then it won't feature so prominently, and this placement is more or less the same as the similar orphan warning boxes of RMS stratigraphy and masterdata editing.

GibranAlfa:
Addressed feedbacks by renaming well to wellbore when appropriate and showing orphan warning before the tips text.

Also fixed a regression introduced by upgrading the heyapi/openapi-ts. After the upgrade, opening a page that checks SMDA (like masterdata, stratigraphy page) now redirects directly to SSO, even when the user only wants to view the page in read-only mode.

After updating @hey-api/openapi-ts, the generated Axios runtime now builds the complete URL before calling Axios and passes an empty baseURL. As a result, the interceptor receives an absolute URL instead of the relative URL used before. Our exact relative path comparison no longer recognizes the SMDA health check. A 401 response is therefore treated as a normal SMDA request and triggers SSO.

The fix compares the normalized URL pathname instead. This supports both URL formats, prevents the health check from triggering SSO, and keeps the existing token refresh behavior for actual SMDA requests.

joargr:
> The fix compares the normalized URL pathname instead. This supports both URL formats, prevents the health check from triggering SSO, and keeps the existing token refresh behavior for actual SMDA requests.

That's a good fix!

frontend/src/components/project/rms/Wellbores.tsx
```diff
@@ -0,0 +1,690 @@
...
+          wellbores shown by the filter.
+        </List.Item>
+        <List.Item>
+          Use Include checkboxes to select or deselect individual wellbores.
+        </List.Item>
+        <List.Item>
+          Mark a wellbore as planned to store it without making it available for
+          wellbore mapping.
+        </List.Item>
+      </PageList>
```
joargr: Suggest some rephrasing of these tips:

```suggestion
      <PageText $marginBottom="0">💡 Tips:</PageText>
      <PageList>
        <List.Item>
          When there are no wellbores stored in the project, all available RMS
          wellbores in the list are initially selected
        </List.Item>
        <List.Item>
          The list can be filtered by wellbore name, and <i>Select</i> and{" "}
          <i>Deselect</i> buttons will then operate on the filtered list
        </List.Item>
        <List.Item>
          Use the <i>Include</i> checkboxes to select individual wellbores for
          storing to the project
        </List.Item>
        <List.Item>
          Mark a wellbore as planned to store it in the project without making
          it available for wellbore mapping
        </List.Item>
      </PageList>
```

The rephrased texts should hopefully be even clearer. Also, there are no periods at the end of each list item, similar to text items in other lists in the application.

frontend/src/components/project/rms/Wellbores.tsx
```diff
@@ -0,0 +1,690 @@
...
+            orphanWellboreNames.length === 1 ? "is" : "are"
+          } currently not available in RMS. ${
+            orphanWellboreNames.length === 1 ? "It" : "They"
+          } will be removed when you save.`}
+          listItems={orphanWellboreNames}
+        />
+      )}
+
+      <PageText $marginBottom="0">💡 Tips</PageText>
+      <PageList>
```
joargr: To reduce the spacing between the tips list and the action buttons, the bottom margin of `PageList` should be set to 0. To do this the component needs to have the `$marginBottom` property, which can be defined as follows:

```ts
export const PageList = styled(List).attrs<{
  $marginBottom?: string;
}>((props) => ({
  $marginBottom: props.$marginBottom ?? tokens.spacings.comfortable.medium,
}))`
  margin-bottom: ${({ $marginBottom }) => $marginBottom};
`;
```

frontend/src/components/project/rms/Wellbores.tsx
```diff
@@ -0,0 +1,690 @@
...
+          native buttons/inputs without an explicit type, so having them inside
+          a form would cause accidental submits when sorting or filtering.
+        */}
+        <form
+          onSubmit={(e) => {
+            e.preventDefault();
+            e.stopPropagation();
+            void form.handleSubmit();
+          }}
+        >
```
joargr: Suggest having `<form>` inside `<Dialog.Actions>`. It seems that there is some CSS rules in EDS that adds padding when `<Dialog.Actions>` is the first child. Ensuring this component is a child of `<Dialog>` (`EditDialog`) and not `<form>` removes some extraneous padding.

GibranAlfa:
Changes after review:

- Rephrased the wellbore tips and removed the list item periods.
- Removed the extra spacing below the tips list.
- Moved the form inside Dialog.Actions to remove the extra EDS padding.

PR #484, opened 2026-07-20
DEV: Update ESlint

### LINKED ISSUE #472: Update ESLint to version 10
The version of ESLint currently used in the application is 9.39.2. Version 9.x end-of-life is 2026-08-06, so ESLint should be updated to version 10.x (currently 10.5.0) before that.

nourinmohd:
Resolves #472 

This PR updates the frontend lint toolchain from ESLint 9 to ESLint 10 ahead of the ESLint 9 end-of-life date.

Changes made : 
- Upgraded `eslint` to `^10.5.0`.
- Upgraded `@eslint/js` to `^10.0.1`.
- Updated related `ESLint plugins/config` packages to versions compatible with `ESLint 10`:
`typescript-eslint`, `@stylistic/eslint-plugin,` `TanStack lint plugins`, `React Hooks`, `React Refresh`, `React X`, and `React DOM lint plugins`.
- Refreshed `pnpm-lock.yaml `for the new dependency graph.
- Updated `eslint.config.js` to keep the current lint contract stable after the preset upgrades.
- Explicitly disabled newly introduced preset rules that flagged existing code patterns:
`react-hooks/refs, react-hooks/set-state-in-effect, react-x/set-state-in-effect, react-x/use-state, no-useless-assignment, and react-refresh/only-export-components.` 

Important Note: 
Some of the newest compatible package releases were inside the repository’s `minimumReleaseAge` quarantine window, so the upgrade uses the latest compatible versions that `pnpm` accepts under the current workspace policy.

## Checklist


frontend/eslint.config.js
```diff
@@ -48,10 +48,15 @@ export default tseslint.config(
...
-        "warn",
+        "off",
         { allowConstantExport: true },
       ],
       ...reactX.configs["recommended-typescript"].rules,
+      "react-hooks/refs": "off",
+      "react-hooks/set-state-in-effect": "off",
+      "react-x/set-state-in-effect": "off",
+      "react-x/use-state": "off",
+      "no-useless-assignment": "off",
```
joargr: I think this rule is useful, so it should be enabled. And more so as it is a recommended rule. The code it finds shouldn't be too difficult to fix.

frontend/package.json
```diff
@@ -32,24 +32,24 @@
...
-    "eslint": "^9.39.4",
-    "eslint-plugin-react-dom": "^2.7.2",
-    "eslint-plugin-react-hooks": "^5.2.0",
-    "eslint-plugin-react-refresh": "^0.4.20",
-    "eslint-plugin-react-x": "^1.49.0",
+    "eslint": "^10.5.0",
+    "eslint-plugin-react-dom": "^5.14.0",
+    "eslint-plugin-react-hooks": "^7.1.0",
+    "eslint-plugin-react-refresh": "^0.5.2",
+    "eslint-plugin-react-x": "^5.14.0",
```
joargr, line 49: Suggest using all plugins from the ESLint React project, and not just the `eslint-plugin-react-dom` and `eslint-plugin-react-x` plugins. There is a now an [issue](https://github.com/equinor/fmu-settings-gui/issues/494) for this, so it can be implemented separately.

frontend/eslint.config.js
```diff
@@ -48,10 +48,15 @@ export default tseslint.config(
         { blankLine: "always", prev: "*", next: "return" },
       ],
       "react-refresh/only-export-components": [
-        "warn",
+        "off",
         { allowConstantExport: true },
       ],
       ...reactX.configs["recommended-typescript"].rules,
+      "react-hooks/refs": "off",
```
joargr: This rule shouldn't be disabled. It triggers on one instance, on the component `LockStatus` in `components/LockStatus.tsx`. It would be better to fix this `ref` error, as part of this PR. The example of `Popover` in the [EDS documentation](https://storybook.eds.equinor.com/?path=/docs/data-display-popover--docs) (by using `useState`) is a fix that will work.

frontend/eslint.config.js
```diff
@@ -48,10 +48,15 @@ export default tseslint.config(
...
+      "react-hooks/set-state-in-effect": "off",
```
joargr: This rule shouldn't be disabled. As there are a number of errors reported from this the rule can stay disabled in this PR. There is instead a separate [issue](https://github.com/equinor/fmu-settings-gui/issues/495) for fixing these errors.

frontend/eslint.config.js
```diff
@@ -48,10 +48,15 @@ export default tseslint.config(
...
+      "react-x/set-state-in-effect": "off",
```
joargr: The same comment as for the `react-hooks/set-state-in-effect` applies for this rule.

frontend/eslint.config.js
```diff
@@ -48,10 +48,15 @@ export default tseslint.config(
...
+      "react-x/use-state": "off",
```
joargr: This rule shouldn't be disabled. The warnings currently reported from this rule should be simple to fix, and could be done in this PR.

frontend/eslint.config.js
```diff
@@ -48,10 +48,15 @@ export default tseslint.config(
         { blankLine: "always", prev: "*", next: "return" },
       ],
       "react-refresh/only-export-components": [
-        "warn",
+        "off",
```
joargr: It seems that the new version of this rule (as of 0.5.0) now is triggering warnings. It makes sense to temporarily disable the rule in this PR. An [issue](https://github.com/equinor/fmu-settings-gui/issues/496) has been created for changing the code so that Hot Module Replacement can work properly in these cases. When this issue has been implemented, the rule can be enabled again.

GibranAlfa:
Addressed the feedbacks by reenabling react-hooks/refs, react-x/use-state, and no-useless-assignment, and fix ref error in LockStatus.

frontend/eslint.config.js
```diff
@@ -47,10 +47,12 @@ export default tseslint.config(
...
         { allowConstantExport: true },
       ],
       ...reactX.configs["recommended-typescript"].rules,
+      "react-hooks/set-state-in-effect": "off",
+      "react-x/set-state-in-effect": "off",
```
joargr: With the most recent updates it seems that these two rules now no longer trigger on the code. Version 7.1.0 of `eslint-plugin-react-hooks` [talks](https://github.com/react/react/blob/main/packages/eslint-plugin-react-hooks/CHANGELOG.md#710) of an improvement of the `set-state-in-effect` rule, with fewer false negatives, so perhaps this affects the linting of the code? These two rules could thus be enabled again, by deleting the disabling lines.

The mentioned [issue](https://github.com/equinor/fmu-settings-gui/issues/495) has been updated and changed to be an issue about general refactoring of `useEffect` code.

GibranAlfa: Good catch, I enabled the two rules again

frontend/eslint.config.js
joargr, line 1: I've committed a refactored version of the configuration file. The changes are based on the recommended format as documented in the newest version of ESLint as well as the plugins that are used. The changes are as follows:
- Uses `defineConfig` from ESLint instead of `config` from `typescript-eslint`
- The configuration contained two configuration objects, so the parameter to `defineConfig` is now an array, which now contains four configuration objects (due to other changes). All objects are merged into one configuration when ESLint is run
- The first configuration object contained only `ignores`, which means that the ignores applied to all configuration objects. The `globalIgnores` helper is now used to make this more readable
- The order of the configuration object properties now follows the order of the [documentation](https://eslint.org/docs/latest/use/configure/configuration-files)
- The TanStack Query plugin has been changed to use the strict ruleset, which are the recommended rules plus some extra rules. The extra rules do not trigger errors or warnings on the code
- Some plugins produce rules lists suitable for inclusion in `extends`, so no need to specify these in `plugins` and `rules`
- Some plugins produce a configuration object, so these are now specified as such
- There are no changes to the ESLint Stylistic configuration, but the `plugin` specification has been modified to mimic its documentation
- Configuration for the `no-unused-vars` has now been placed directly before the `@typescript-eslint/no-unused-vars` rule, as these two rules need to be configured in tandem
- The `reactRefresh` plugin can produce a configuration object for use with Vite, which sets the  `allowConstantExport` option. There is thus no need to set this option specifically
- As the `react-refresh/only-export-components` rule is to be temporarily disabled, and its plugin now produces a configuration object, the configuration is temporarily modified so as to override and disable this particular rule. The plugin only contains this one rule, but this temporary configuration syntax doesn't change any other rules, should such become available. Once the [issue](https://github.com/equinor/fmu-settings-gui/issues/496) to fix the code has been implemented, the configuration can be changed to the simple, non-overridden configuration object, which is currently commented out

joargr, line 1: An additional comment is that when a configuration object does not include a `files` (or `ignores`) property, the rules in that object are applied to the files [matched by any other configuration objects](https://eslint.org/docs/latest/use/configure/configuration-files#specify-files-and-ignores). That means that the `files` defined in the largest object also apply for the configuration objects produced by the plugins.

PR #485, opened 2026-07-20
DEV: Update @hey-api/openapi-ts

### LINKED ISSUE #471: Update @hey-api/openapi-ts to latest version
[@hey-api/openapi-ts](https://heyapi.dev/) is used to generate a client and functions to communicate with the FMU Settings API, based on the `openapi.json` API specification. It is currently at version 0.66.7 (released April 2025), and should be updated to the latest version. After the update, the API code should be regenerated, and the application code adjusted as needed.

There have been a significant number of releases since the version currently in use, so there are breaking changes that need to be dealt with. A few important ones are the following:

- Client plugins are now [bundled with the main package](https://heyapi.dev/docs/openapi/typescript/migrating#v0730), so the current dependency of `@hey-api/client-axios` is no longer needed, and should be removed from `package.json`
- `POST` operations are now treated as [mutations and not queries](https://github.com/hey-api/hey-api/releases/tag/%40hey-api%2Fopenapi-ts%400.82.0). Some FMU Settings `POST` operations are used as queries (such as searches), so there needs to be configuration to enable this. Suggest using something like the following (additional paths should be added to the list, as needed):

```ts
  parser: {
    hooks: {
      operations: {
        isQuery: (op) => {
          if (op.method === "post" && ["/api/v1/match"].includes(op.path)) {
            return true;
          }
        },
      },
    },
  },
```
- There are changes to how formatting and linting of the generated code is configured. Given the way `biome` is installed, the commands should be configured as follows:

```ts
  output: {
    path: "src/client",
    postProcess: [
      {
        command: "tools/biome",
        args: ["format", "--config-path=../biome.json", ".."],
      },
      {
        command: "tools/biome",
        args: ["lint", "--config-path=../biome.json", ".."],
      },
    ],
  },
```

There is the ability to [store](https://heyapi.dev/docs/openapi/typescript/configuration/output#source) the `openapi.json` API specification used when generating the code. This could be an interesting feature to have, as it shows exactly what the input for the generated code is, and it can be useful when reviewing PRs as it clearly shows the changes in the API that is part of the PR.

nourinmohd:
Resolves #471 

This PR updates the frontend OpenAPI client generation from `@hey-api/openapi-ts@0.66.7` to `@hey-api/openapi-ts@0.99.0 `and regenerates the API client from the current FMU Settings API OpenAPI specification.

Main changes:

- Updated `package.json` and `pnpm-lock.yaml`:
  - `@hey-api/openapi-ts` is now `^0.99.0`.
  - Removed the direct `@hey-api/client-axios` dependency, because client plugins are now bundled with `@hey-api/openapi-ts`.
  

- Updated `openapi-ts.config.ts:`

    - Replaced old `output.format / output.lint` config with `output.postProcess.`
    - Configured post-processing to run the repo-local tools/biome formatter and linter.
    - Added `output.source` so the OpenAPI input is stored at `openapi.json`.
    - Added `parser.hooks.operations.isQuery` so selected POST endpoints still generate TanStack Query query options:
    ```
    /api/v1/match
    /api/v1/smda/field
    /api/v1/smda/masterdata
    /api/v1/smda/strat_units
    ```
- Regenerated generated client code under client:

- Existing generated files changed due to the new generator version.
    - New generated runtime folders were added:
    `client`
    `core`
    `source`
    - These should be reviewed as generated output, not handwritten application logic.

- Updated `tsconfig.node.json`:
  - Added DOM libs because the new generated `client/runtime` types reference browser APIs through the config typecheck path.
- Updated `README.md`:
  - Documented that the OpenAPI source snapshot is tracked with the generated client.

## Checklist


frontend/openapi-ts.config.ts
```diff
@@ -4,8 +4,39 @@ export default defineConfig({
...
+    hooks: {
+      operations: {
+        isQuery: (op) => {
+          if (
+            op.method === "post" &&
+            [
+              "/api/v1/match",
+              "/api/v1/smda/field",
+              "/api/v1/smda/masterdata",
+              "/api/v1/smda/strat_units",
```
GibranAlfa, line 39: add /api/v1/smda/well_headers as well to search for well names in SMDA

frontend/tsconfig.node.json
joargr, line 1: This file is not relevant, as the application only runs in the browser. As such, I don't think there should be done any updates to the file.

GibranAlfa: good catch, the changes now unnecessary after we added @ts-nocheck to the generated client/ and core/ files, so i revert the update

PR #488, opened 2026-07-21
ENH: Add changelog page

### LINKED ISSUE #456: Display the full changelog in the history page
Description to be added after discussion with the team.

nourinmohd:
Resolves #456

This PR adds a dedicated full changelog page for selected projects while keeping the homepage changelog as a compact “recent changes” preview.

Main Changes
- Added a new project route: `changelog.tsx`. 
    - Shows the full changelog for the selected project.
    - Handles the no-project-selected state.
    - Includes a short intro explaining the page and filters.

- Added `Changelog` to the Project sidebar in `Sidebar.tsx`, under `History`.

- Updated `Changelog.tsx`.
    - Homepage changelog still shows only the most recent entries.
    - Adds a link from the homepage preview to the full changelog page.
    - Full changelog now renders as an `EdsDataGrid`.
    - Table supports filtering by `change type` and `number of entries`.
    - Table uses full page width via `PageContainerNotWidthConstrained`.
    - Each row has an outlined `View details` button.
 
- Improved changelog detail rendering in `utils.ts`. 
    - Parses `Old value / New value `changelog payloads.
    - Attempts to extract only the changed nested field using `entry.key / entry.path`.
    - Falls back to the original full value if parsing or path extraction fails.

- Added styling in `Changelog.style.ts`. 
    - Full-width table container.
    - Filter controls.
    - Details dialog layout.
    - Colored before/after panels inspired by the cache diff view.   

## Checklist

### DIFF
--- a/frontend/src/components/Sidebar.tsx
+++ b/frontend/src/components/Sidebar.tsx
@@ -89,6 +89,7 @@ export function Sidebar() {
       ],
     });
     ProjectSubItems.push({ label: "History", to: "/project/history" });
+    ProjectSubItems.push({ label: "Changelog", to: "/project/changelog" });
   }
 
   return (
--- a/frontend/src/components/home/Changelog.style.ts
+++ b/frontend/src/components/home/Changelog.style.ts
@@ -27,6 +27,69 @@ export const ChangeList = styled.div`
   gap: ${tokens.spacings.comfortable.small};
 `;
 
+export const ChangelogHeader = styled.div`
+  display: flex;
+  align-items: baseline;
+  justify-content: space-between;
+  gap: ${tokens.spacings.comfortable.small};
+  margin-bottom: ${tokens.spacings.comfortable.small};
+`;
+
+export const ChangelogGroup = styled.section`
+  & + & {
+    margin-top: ${tokens.spacings.comfortable.large};
+  }
+`;
+
+export const ChangelogGroupHeader = styled.h3`
+  margin: 0 0 ${tokens.spacings.comfortable.small};
+  color: ${tokens.colors.text.static_icons__secondary.hex};
+  font-size: 0.875rem;
+  font-weight: 500;
+`;
+
+export const ChangelogTableContainer = styled.div`
+  height: 70vh;
+  min-height: 30rem;
+  overflow: auto;
+
+  .table-wrapper {
+    height: 100%;
+  }
+
+  table {
+    width: 100% !important;
+  }
+`;
+
+export const ChangelogFilterBar = styled.div`
+  display: flex;
+  align-items: flex-end;
+  gap: ${tokens.spacings.comfortable.small};
+  flex-wrap: wrap;
+  margin-bottom: ${tokens.spacings.comfortable.medium};
+`;
+
+export const ChangelogFilterField = styled.label`
+  display: flex;
+  flex-direction: column;
+  gap: ${tokens.spacings.comfortable.xx_small};
+  min-width: 12rem;
+  color: ${tokens.colors.text.static_icons__default.hex};
+  font-size: 0.875rem;
+  font-weight: 500;
+`;
+
+export const ChangelogFilterSelect = styled.select`
+  min-height: 2.5rem;
+  padding: 0 ${tokens.spacings.comfortable.small};
+  border: 1px solid ${tokens.colors.ui.background__medium.hex};
+  border-radius: ${tokens.shape.corners.borderRadius};
+  background: ${tokens.colors.ui.background__default.hex};
+  color: ${tokens.colors.text.static_icons__default.hex};
+  font: inherit;
+`;
+
 export const ChangeItem = styled.article<{ $changeType: ChangeType }>`
   padding: ${tokens.spacings.comfortable.small} ${tokens.spacings.comfortable.medium};
   border-left: 3px solid ${({ $changeType }) => changeTypeColor($changeType)};
@@ -39,6 +102,7 @@ export const ChangeItemHeader = styled.div`
   align-items: center;
   justify-content: space-between;
   gap: ${tokens.spacings.comfortable.small};
+  flex-wrap: wrap;
 `;
 
 export const ChangeTypeChip = styled(Chip)<{ $changeType: ChangeType }>`
@@ -59,4 +123,70 @@ export const ChangeItemMeta = styled.div`
   margin-top: ${tokens.spacings.comfortable.xx_small};
   color: ${tokens.colors.text.static_icons__secondary.hex};
   font-size: 0.875rem;
+`;
+
+export const ChangeItemField = styled(ChangeItemMeta)`
+  overflow-wrap: anywhere;
+`;
+
+export const ChangeDetails = styled.div`
+  margin-top: ${tokens.spacings.comfortable.small};
+  padding: ${tokens.spacings.comfortable.small};
+  border: 1px solid ${tokens.colors.ui.background__medium.hex};
+  border-radius: ${tokens.shape.corners.borderRadius};
+  background: ${tokens.colors.ui.background__default.hex};
+`;
+
+export const ChangeDetailsHeader = styled.div`
+  display: flex;
+  align-items: flex-start;
+  justify-content: space-between;
+  gap: ${tokens.spacings.comfortable.small};
+  margin-bottom: ${tokens.spacings.comfortable.small};
+`;
+
+export const ChangeDetailsSummary = styled.div`
+  margin-bottom: ${tokens.spacings.comfortable.small};
+  color: ${tokens.colors.text.static_icons__secondary.hex};
+`;
+
+export const ChangeDetailsValueGrid = styled.div`
+  display: grid;
+  grid-template-columns: repeat(2, minmax(0, 1fr));
+  gap: ${tokens.spacings.comfortable.small};
+
+  @media (max-width: 48em) {
+    grid-template-columns: 1fr;
+  }
+`;
+
+export const ChangeDetailsValuePanel = styled.div<{
+  $kind: "before" | "after";
+}>`
+  padding: ${tokens.spacings.comfortable.small};
+  border: 1px solid
+    ${({ $kind }) =>
+      $kind === "before"
+        ? tokens.colors.interactive.danger__resting.hex
+        : tokens.colors.interactive.success__resting.hex};
+  border-radius: ${tokens.shape.corners.borderRadius};
+  background: ${({ $kind }) =>
+    $kind === "before"
+      ? tokens.colors.ui.background__danger.hex
+      : tokens.colors.interactive.success__highlight.hex};
+`;
+
+export const ChangeDetailsValueHeader = styled.div`
+  margin-bottom: ${tokens.spacings.comfortable.small};
+  font-weight: 500;
+`;
+
+export const ChangeDetailsContent = styled.pre`
+  margin: 0;
+  overflow-x: auto;
+  white-space: pre-wrap;
+  overflow-wrap: anywhere;
+  color: ${tokens.colors.text.static_icons__default.hex};
+  font-size: 0.875rem;
+  line-height: 1.4;
 `;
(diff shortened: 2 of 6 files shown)


nourinmohd:
Screenshot 2026-07-21 at 09 37 16
Screenshot 2026-07-21 at 09 37 33
Screenshot 2026-07-21 at 09 37 48
Screenshot 2026-07-21 at 09 38 00

GibranAlfa (commented):
This PR looks very rushed. You need to think more about the design and understand the code.

### Homepage preview

- “View full changelog” looks disconnected below the list. I think it should be placed directly after the text explaining that the five most recent changes are shown.
- The homepage should request `max_entries=5` instead of fetching the complete changelog and slicing it in the GUI.

### Component structure

The full changelog feature should not live in `components/home/Changelog.tsx`.

I suggest keeping the homepage preview there and moving the full-page implementation into a feature folder, for example:

- `components/project/changelog/Changelog.tsx`
- `components/project/changelog/ChangelogTable.tsx`
- `components/project/changelog/ChangelogDetailsDialog.tsx`
- `components/project/changelog/Changelog.style.ts`
- `components/project/changelog/utils.ts`

### Table columns

Current table header looks very technical and not understandable for average users. I suggest simplifying the table to:

| Date | Change | Settings type | Changed by | Details |
|---|---|---|---|---|

- Remove the separate **Change type** column. The Change text already starts with “Updated,” “Added,” “Restored,” etc.
- Remove the **Field** column. Values such as `cache_max_revisions` are technical and duplicate the user-friendly Change description.
- Rename **File** to **Settings type**.
- Rename **User** to **Changed by**.
- `entry.path` should not be used as a fallback for the setting key because it represents the project filesystem path, not the changed setting.
- Some entries still fall back to technical descriptions such as Updated field 'created_at'. If these entries should be visible to users, they should have friendly descriptions.

### Filters

I suggest three visible controls above the table:

- **Filter changes by:** All changes, Modified, Added, Removed, Restored, etc.
- **Settings type:** All settings types, Project configuration, Mappings.
- **Show:** All changes, Latest 10, Latest 25, Latest 50, Latest 100.

These should use an EDS component such as `NativeSelect` instead of custom-styled HTML `<select>` elements.

Settings type is the additional filter. The API supports this by sending this query params for example:
```text
field_name=file
filter_value=config.json
filter_type=text
operator===
```

### Use the API filters

The current implementation fetches the full changelog with `projectGetChangelogOptions()` and then filters and slices the result in the GUI.

Instead, changing the controls should update the request:

- Send `change_type` when a specific change type is selected.
- Send `max_entries` when Latest 10/25/50/100 is selected.
- Omit `max_entries` when All changes is selected.
- Filter Settings type using `field_name=file`, `filter_type=text`, and `operator===`.

The API applies the filters before `max_entries`, so selecting Modified + Project configuration + Latest 25 should return the latest 25 matching entries.

### `EdsDataGrid` layout and performance

- Add `enableVirtual` and pass a fixed `height` directly to `EdsDataGrid`. The full changelog may contain many entries, and virtualization ensures that the grid renders only the rows currently visible in the scrollable area instead of mounting every row at once. This keeps scrolling and filtering responsive as the changelog grows.
- Give the columns intentional `size` and `minSize` values. Currently all columns width look the same without proper thoughts on each column.
- Let Change use most of the available width.
- Sorting could be enabled for Date, Settings type, and Changed by.

### Opening details

There are currently two ways to open the same dialog: clicking the row and clicking “View details.”

I suggest keeping the explicit **View details** button and removing `onRowClick`. The row does not look clickable, and supporting both interactions is redundant.

### Details dialog

The dialog grows based on its content. It should have:

- A maximum content height, for example `70vh`.
- `overflow-y: auto` on the content.
- A header and Close action that remain visible.

### Changelog parser

There are several issues in the current parser:

- The old value still includes `->`. For example, it displays `5 ->` instead of `5`.
- It only recognizes entries containing both `Old value:` and `New value:`.
- Added entries containing only `New value:` are incorrectly shown as if their details were not recorded.
- Removed entries containing only `Old value:` will have the same problem.

The parser should at least handle updates, additions, removals, empty values, and the `-> New value:` delimiter correctly.

### Structured diff

I think the proper long-term fix should start in `fmu-settings`.

- Update `ProjectConfigManager` and `MappingsManager` to call `get_structured_model_diff()` with the resource models from before and after the update, then pass the resulting differences to `log_update_to_changelog()`.
- Store it as an optional field on `ChangeInfo`, preserving compatibility with existing changelog files.
- Change `fmu-settings-api` to expose the newly added structured diff field.
- Let the GUI render the structured difference when available and use the string parser only for existing entries (old changelog without structured diff).

GibranAlfa:
Haven't look too much into the code, but in a first glance:

- Virtualization `enableVirtual` is still not used in `EdsDataGrid`
- Sorting for the columns are not implemented, although this is optional, I think it's nice to have and easy to implement as well as `EdsDataGrid` already supports it

Some other ideas:
- Add a EDS column filter button in "Changed by" column. This will only filter the results in the GUI.
- Maybe better to show the changelog table with pagination. But this requires API changes to support it, otherwise client-side pagination is not really useful now.

frontend/src/components/Sidebar.tsx
```diff
@@ -88,6 +88,7 @@ export function Sidebar() {
       ],
     });
     ProjectSubItems.push({ label: "History", to: "/project/history" });
+    ProjectSubItems.push({ label: "Changelog", to: "/project/changelog" });
```
joargr, line 91: I think the changelog is more relevant for the user than the history page:

- There is a short changelog section on the homepage, so the changelog page is an extension of that
- The history page allows for restoring previous configurations, which I think is a feature that will not be used that often

The changelog page should thus appear before the the history page in the menu.

frontend/src/components/project/changelog/ChangelogTable.tsx
joargr, line 1: I think the "Settings type" column should appear before the "Change" column. The settings type (Project configuration or Mappings) sets the context of the change, so I would say that it's more useful to first know where a change has happened, and then see what the change is. If the columns switch place, the corresponding filter widgets should also switch place.

joargr, line 1: It would be good if the date value isn't wrapped over several lines, as happens when the browser width narrows. Perhaps the date part and the time part each can be prevented from wrapping? That is, with a narrow browser window the only wrapping that occurs is a line break between the date and the time.

If the browser width is equal to the confined layout width (the width of the main page content), the date value will wrap just before "AM"/"PM". Ensuring that neither date nor time will wrap will prevent a wrap on this place. So suggest preventing such wraps, and also suggest adding some width on the Date column so that the full date value (date and time) don't wrap when the main content is shown with the confined layout width.

joargr, line 1: The changelog page uses the full width of the browser. I think this is a powerful layout option which should be used sparingly. Currently only the RMS stratigraphy and stratigraphy mappings pages allow for this, and even then the width is initially constrained. I don't think there is that much data in a changelog table row that a full width is necessary, so suggest that the width be constrained, without an option to expand to full width.

frontend/src/components/project/changelog/ChangelogTable.tsx
```diff
@@ -0,0 +1,82 @@
...
+      cell: ({ row }) =>
+        row.original.timestamp
+          ? displayDateTime(row.original.timestamp)
+          : "(unknown date)",
+    },
+    {
+      id: "description",
+      header: "Change",
+      accessorFn: (entry) => formatEntryDescription(entry),
+    },
```
joargr, line 37: When the width of this column is narrow, the value will be truncated. Suggest that the value be wrapped instead. It seems that the default is for the value to be truncated, but when the display is defined, by `cell`, there is no truncation done:

```ts
      cell: ({ getValue }) => getValue(),
```

frontend/src/components/project/changelog/utils.ts
```diff
@@ -0,0 +1,321 @@
...
+  return undefined;
+}
+
+function formatBriefDescription(entry: ChangeInfo) {
+  const change = entry.change;
+  const compact = change.replace(/\s+/g, " ");
+  const withoutDiffPayload = compact.replace(/\. Old value:.*/, "");
+  const technicalFieldChange = TECHNICAL_FIELD_CHANGE_PATTERN.exec(
+    withoutDiffPayload,
+  );
```
joargr, line 108: This generates a linting error. The proper formatting is the following:

```suggestion
  const technicalFieldChange =
    TECHNICAL_FIELD_CHANGE_PATTERN.exec(withoutDiffPayload);
```

frontend/src/components/project/changelog/ChangelogDetailsDialog.tsx
```diff
@@ -0,0 +1,124 @@
...
+  formatSettingLabel,
+  getTypeLabel,
+  parseChangeDetails,
+} from "./utils";
+
+export function ChangelogDetailsDialog({
+  entry,
+  onClose,
+}: {
+  entry?: ChangeInfo;
```
joargr: This should allow for the value being `undefined`.

```suggestion
  entry?: ChangeInfo | undefined;
```

There is currently a type error reported for the calling of this component in `ChangelogTable`.

frontend/src/components/home/Changelog.tsx
```diff
@@ -43,53 +66,59 @@ function Content() {
...
-  const latestChanges = [...data]
-    .sort((a, b) => (b.timestamp ?? "").localeCompare(a.timestamp ?? ""))
-    .slice(0, 5);
-
   return (
     <>
       <PageText>
-        {latestChanges.length === 1
+        {changes.length === 1
           ? "Showing the most recent change to this project's settings."
```
joargr: There's recently been some rephrasing in the application so that the term being used to refer to the project is "the project" and not "_this_ project". The three occurrences in this file should be changed, as well as the two occurrences in `components/project/changelog/Changelog.tsx`.

frontend/src/components/project/changelog/utils.ts
```diff
@@ -0,0 +1,321 @@
+import type { ChangeInfo, ChangeType } from "#client/types.gen";
+
+export const FILE_LABELS: Record<string, string> = {
+  "config.json": "Project configuration",
+  "mappings.json": "Mappings",
+};
+
+const PATH_LABELS: Record<string, Record<string, string> | undefined> = {
+  "config.json": {
```
joargr: In test data there is a changelog key `validation.rms_project`, which currently results in the changelog entry "Updated rms project in Project configuration". This might not be the best description. It also looks odd when placed next to "Updated RMS wells in Project configuration", as RMS should always be spelled with capital letters.

There should be a description (label) for such a change. I'm not sure what a good description would be, but the value of `change` for the entry is `Updated field 'validation.rms_project'. Old value: None -> New value: None"`.

frontend/src/components/home/Changelog.tsx
```diff
@@ -43,53 +66,59 @@ function Content() {
...
-          : `Showing the ${latestChanges.length} most recent changes to this project's settings.`}
+          : `Showing the ${changes.length} most recent changes to this project's settings.`}{" "}
+        <Typography link as={Link} to="/project/changelog">
+          View full changelog
+        </Typography>
```
joargr: It should be avoided to have a text link like "View full changelog". The link text is an action to do, as used for button links. A button is not appropriate here, so the text and the link should be rephrased so that the sentence has a meaning even without a link. Suggest the following:

```suggestion
        The{" "}
        <Typography link as={Link} to="/project/changelog">
          changelog page
        </Typography>{" "}
        shows all changes.
```

frontend/src/components/home/Changelog.style.ts
joargr: There is no need to set the top margin on this styled component. This margin is [collapsed](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Box_model/Margin_collapsing) with the bottom margin of the element above. In general top margins does not need to be set, but there's often a need to set a bottom margin.

PR #492, opened 2026-07-23
ENH: Replace GUI server with static asset package

GibranAlfa:
Part of equinor/fmu-settings-api#433

Removed the standalone Python GUI server. The Python package now builds and contains the React static files and exposes their location through `get_static_directory()`.

Production frontend requests use the browser origin, while the existing Vite development flow continues to use the API on port 8001. FastAPI and Uvicorn are no longer GUI package dependencies.

## Review order

This PR and [fmu-settings-api#438](https://github.com/equinor/fmu-settings-api/pull/438) can be reviewed in parallel. They define the interfaces used by the CLI.

Both should be merged and released before the CLI PR is finalized. The optional `fmu-settings` documentation PR can be reviewed last.

## Checklist

- [x] Tests added


joargr:
Now that the backend only runs one server (for API and GUI files), on port 8000, I guess the frontend (GUI) should use the same server (port 8000) also for development, instead of using port 8001. In other words, the openapi-ts and API configuration as well as the documentation and examples in the various repos should be updated so that only the 8000 port is used.

GibranAlfa:
> Now that the backend only runs one server (for API and GUI files), on port 8000, I guess the frontend (GUI) should use the same server (port 8000) also for development, instead of using port 8001. In other words, the openapi-ts and API configuration as well as the documentation and examples in the various repos should be updated so that only the 8000 port is used.

Yes we can simplify this, so API will only be served in port 8000.

